PR #11, opened 2021-07-09
Test module00a

pdiakumis:
@illusional this seems to be running fine with the https imports. Is a tag of our fork okay (https://github.com/populationgenomics/gatk-sv/tree/v0.14-beta)?


illusional (approved):
Yeah, looks great @pdiakumis pet

PR #21, opened 2021-08-04
Use skopeo for image copying to AR

pdiakumis:
Will run with:

```
analysis-runner \
    --dataset "fewgenomes" \
    --output-dir "pdiakumis/sv-workflows/docker" \
    --description "sync docker images to AU AR" \
    --access-level "standard" \
    docker_sync.R
```


vladsaveliev (approved):
Awesome stuff!

But why not use docker?

PR #26, opened 2021-08-24
Gather sample SV evidence per batch

pdiakumis:
This WDL workflow (from GATK-SV v0.17.1-beta) takes as input arrays of CRAMs (and sample IDs) and runs Manta/Delly/Wham, PESRCollection, CollectCounts, and GatherSampleEvidenceMetrics from GATK-SV. Outputs can then be used for TrainGCNV which trains a gCNV model (separate PR).
Tested on TOB1520/21/22 during August 23/24, and had a runtime of ~13 hours.


vladsaveliev (approved):
I'm genuinely impressed with your patience, the number of dockers alone O_o

PR #38, opened 2022-02-16
Hoptan

hopedisastro:
Please pull `copy_gangstr_image.sh`. GangSTR is a short-read STR caller which will be used in future workflows.


str/batch.py
```diff
@@ -0,0 +1,75 @@
...
+    # The command that do the actual job.
+    j.command(
+        f"""
+    samtools view -T {ref.base} {cram['cram']} {region} -Ocram -o {j.output_bam}
+    """
+    )
+
+    # Speciying where to write the result
+    out_fname = os.path.splitext(os.path.basename(cram_path))[0] + '-split.cram'
+    output_path = f'gs://{HAIL_BUCKET}/{OUTPUT_SUFFIX}/{out_fname}'
```
MattWellie, line 66: there is a [neat helper method](https://github.com/populationgenomics/rare-disease/blob/initial_content/reanalysis/reanalysis_wrapper.py#L14) in analysis-runner for this

illusional: Whatt!!! I can't believe I didn't know about this.

str/gangstr_runner.py
```diff
@@ -0,0 +1,81 @@
...
+            base=REF_FASTA,
+            fai=REF_FASTA + '.fai',
+            dict=REF_FASTA.replace('.fasta', '').replace('.fna', '').replace('.fa', '')
+            + '.dict',
+        )
+    )
+
+    # This image contains basic bioinformatics tools like samtools, bcftools, Picard, etc.
+    j.image(SAMTOOLS_IMAGE)
+    j.image(GANGSTR_IMAGE)
```
illusional, line 51: You can't specify two images, but we'll create a second job that uses this gangstr image.

```suggestion
```

str/gangstr_runner.py
```diff
@@ -0,0 +1,81 @@
...
+    # The command that do the actual job.
+    j.command(
+        f"""
+
+    samtools view -T {ref.base} {cram['cram']} {region} -Obam -o {j.output_bam}    
+    samtools index {j.output_bam}
+    gangstr --bam {j.output_bam} --ref ref.base --regions {regions} --out {j.gangstr_trial}
+    """
+    )
+
```
illusional, line 69: We want to run this gangstr command in a second job, eg:

```suggestion
    """
    )
    gangstr_job = b.new_job("gangstr")
    gangstr.image(GANGSTR_IMAGE)
    # Hail batch recognises the dependency on `j.output_bam`
    gangstr.command(f"""
    gangstr --bam {j.output_bam} --ref {ref.base} --regions {regions} --out {gangstr_job.gangstr_trial}
    """
    )

```

str/gangstr_runner.py
```diff
@@ -0,0 +1,81 @@
...
+    # Speciying where to write the result
+    out_fname = os.path.splitext(os.path.basename(cram_path))[0] + '-split.gang'
+    output_path = f'gs://{HAIL_BUCKET}/{OUTPUT_SUFFIX}/{out_fname}'
+    b.write_output(j.gangstr_trial, output_path)
```
illusional, line 73:
```suggestion
    b.write_output(gangstr_job.gangstr_trial, output_path)
```

str/gangstr_runner.py
```diff
@@ -0,0 +1,81 @@
...
+
+    # don't wait for the hail batch workflow to complete, otherwise
+    # the workflow might get resubmitted if this VM gets preempted.
+    b.run(wait=False)
+
+
+if __name__ == '__main__':
+    main()  # pylint: disable=no-value-for-parameter
```
illusional, line 81: We like a newline character at the end of files :)
```suggestion
    main()  # pylint: disable=no-value-for-parameter
    
```

str/gangstr_runner.py
```diff
@@ -0,0 +1,81 @@
...
+REF_FASTA = 'gs://cpg-reference/hg38/v1/Homo_sapiens_assembly38.fasta'
+SAMTOOLS_IMAGE = 'australia-southeast1-docker.pkg.dev/cpg-common/images/samtools:v0'
+GANGSTR_IMAGE = "australia-southeast1-docker.pkg.dev/cpg-common/images/gangstr:v2.5"
+regions = 'gs://cpg-fewgenomes-test/hoptan-str/gangstr_catalog_with_offtarget.bed'
+
+
+@click.command()
+@click.argument('cram_path')
+@click.argument('region')
+def main(cram_path: str, region: str):  # pylint: disable=missing-function-docstring
```
illusional, line 25: Perform the same treatment as the cram:
```suggestion
@click.argument('region_path')
def main(cram_path: str, region_path: str):  # pylint: disable=missing-function-docstring
```

str/gangstr_runner.py
```diff
@@ -0,0 +1,81 @@
...
+    # Initializing Batch
+    backend = hb.ServiceBackend(billing_project=BILLING_PROJECT, bucket=HAIL_BUCKET)
+    b = hb.Batch(backend=backend, default_image=os.getenv('DRIVER_IMAGE'))
+
+    # Adding a job and giving it a descriptive name.
+    j = b.new_job('Subset CRAM')
+
+    # Making sure Hail Batch would localize both CRAM and the correponding CRAI index
+    cram = b.read_input_group(**{'cram': cram_path, 'cram.crai': cram_path + '.crai'})
+
```
illusional, line 38: As @MattWellie mentioned, you'll want to load the region file in:

```suggestion
region = b.read_input(region_path)
```

illusional (commented):
Some suggestions to run gangstr in a second job, and load the region file into batch.

str/batch.py
MattWellie: I assumed you wrote it?!

MattWellie:
Other than @illusional's suggestions which are all great, my only other suggestion would be (if the analysis script is still being revised) separating the skopeo script onto a clean branch to merge now, and then using the container to develop the analysis script

str/gangstr_runner.py
hopedisastro: Hi Michael, does "region" refer to the second input argument? EG "chr21:1-1000" in analysis-runner \
  --access-level test \
  --dataset fewgenomes \
  --description "Run Batch" \
  --output-dir "$(whoami)-test-batch" \
  batch.py \
  gs://cpg-fewgenomes-test/analysis-runner-test/batch/input.cram \
  chr21:1-10000

str/gangstr_runner.py
hopedisastro: There are two 'region-like' variables. I should change their names so it's more clear. One refers to the variant catalog, which is a bed file. The other is the second command line argument specifying a subset of the chromosome region if needed.

PR #56, opened 2023-01-11
GangSTR Runner

hopedisastro:
Modified https://github.com/populationgenomics/sv-workflows/blob/main/str/str_iterative_eh_runner.py to create a runner for GangSTR. Pylint flagged that there was a lot of duplicate code, but I silenced this warning to pass the check.

### DIFF
--- a/str/str_iterative_gangstr_runner.py
+++ b/str/str_iterative_gangstr_runner.py
@@ -28,7 +28,7 @@ from cpg_utils.hail_batch import remote_tmpdir, output_path, reference_path
 config = get_config()
 
 SAMTOOLS_IMAGE = config['images']['samtools']
-GANGSTR_IMAGE = config ['images']['gangstr']
+GANGSTR_IMAGE = config['images']['gangstr']
 
 
 # inputs:
--- a/str/str_iterative_gangstr_runner.py
+++ b/str/str_iterative_gangstr_runner.py
@@ -3,9 +3,9 @@
 
 """
 This script uses GangSTRv2.5 to call STRs on WGS cram files.
-Required input: --variant-catalog (file path to variant catalog), --project-id, and external sample IDs
+Required input: --variant-catalog (file path to variant catalog), --dataset, and external sample IDs
 For example:
-analysis-runner --access-level test --dataset tob-wgs --description 'tester' --output-dir 'tester' str_iterative_gangstr_runner.py --variant-catalog=gs://cpg-tob-wgs-test/hoptan-str/Illuminavariant_catalog.json --project-id=tob-wgs TOB1XXXX TOB1XXXX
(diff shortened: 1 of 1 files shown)


str/str_iterative_gangstr_runner.py
```diff
@@ -0,0 +1,147 @@
...
+@click.option('--project-id', help='project-id eg tob-wgs')
+# input sample ID
+@click.argument('external-wgs-ids', nargs=-1)
+@click.command()
+def main(
+    variant_catalog, project_id, external_wgs_ids: list[str]
+):  # pylint: disable=missing-function-docstring
+    # Initializing Batch
+    backend = hb.ServiceBackend(
+        billing_project=get_config()['hail']['billing_project'],
```
illusional, line 55:
```suggestion
        billing_project=config['hail']['billing_project'],
```

illusional (changes requested):
Mostly minor changes, hopefully should be straight-forward to address :)

str/str_iterative_gangstr_runner.py
```diff
@@ -0,0 +1,147 @@
...
+
+    external_id_to_cpg_id: dict[str, str] = SampleApi().get_sample_id_map_by_external(
+        project_id, list(external_wgs_ids)
+    )
+    cpg_id_to_external_id = {
+        cpg_id: external_wgs_id
+        for external_wgs_id, cpg_id in external_id_to_cpg_id.items()
+    }
+    if project_id == 'tob-wgs':
+        ref_fasta = (
```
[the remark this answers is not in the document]
vladsavelyev: That would make sense. (Though the discussion on what reference to record in the CRAM header is still ongoing:
https://github.com/populationgenomics/production-pipelines/issues/208)

str/str_iterative_gangstr_runner.py
```diff
@@ -0,0 +1,139 @@
...
+# inputs:
+# variant catalog
+@click.option('--variant-catalog', help='Full path to Illumina Variants catalog')
+# input project ID
+@click.option('--dataset', help='dataset to operate on, eg: tob-wgs')
+# input sample ID
+@click.argument('external-wgs-ids', nargs=-1)
+@click.command()
+def main(
+    variant_catalog, project_id, external_wgs_ids: list[str]
```
illusional, line 43: This project_id and subsequent usages also need to be switched

str/str_iterative_gangstr_runner.py
```diff
@@ -0,0 +1,139 @@
...
+
+SAMTOOLS_IMAGE = config['images']['samtools']
+GANGSTR_IMAGE = config['images']['gangstr']
+
+
+# inputs:
+# variant catalog
+@click.option('--variant-catalog', help='Full path to Illumina Variants catalog')
+# input project ID
+@click.option('--dataset', help='dataset to operate on, eg: tob-wgs')
```
illusional, line 38: Optional: You could make the default here the default from the config you had before. So running the script without a dataset param automatically uses the dataset provided in the config?

illusional (approved):
One small rename of the project_id is required, but otherwise looks good to me

PR #59, opened 2023-01-16
Scripts to run TRTools function

hopedisastro:
This PR includes scripts that run various functions in TRTools package - compareSTR, mergeSTR, qcSTR, and statSTR.

mergeSTR_prep.py is a script that prepares raw VCF output files and makes them compatible for mergeSTR.


str/trtools/merge_str_prep.py
```diff
@@ -0,0 +1,143 @@
...
+def main(
+    dataset, caller, input_dir, external_wgs_ids: list[str]
+):  # pylint: disable=missing-function-docstring
+
+    # Initializing Batch
+    backend = hb.ServiceBackend(
+        billing_project=get_config()['hail']['billing_project'],
+        remote_tmpdir=remote_tmpdir(),
+    )
+    b = hb.Batch(backend=backend, default_image=os.getenv('DRIVER_IMAGE'))
```
MattWellie, line 47: similar to above, you can sub this out for `get_batch` and save yourself a few lines

str/trtools/merge_str_prep.py
```diff
@@ -0,0 +1,143 @@
...
+
+config = get_config()
+
+REF_FASTA = 'gs://cpg-common-main/references/hg38/v0/Homo_sapiens_assembly38.fasta'
+BCFTOOLS_IMAGE = config['images']['bcftools']
+
+
+# inputs:
+# caller
+@click.option('--caller', help='gangstr or eh')
```
MattWellie, line 30:
```suggestion
@click.option('--caller', help='gangstr or eh', type=click.Choice(['eh', 'gangstr'], case_sensitive=True))
```

str/trtools/merge_str_prep.py
```diff
@@ -0,0 +1,143 @@
...
+    input_vcf_dict = {}
+
+    if caller == 'eh':
+        for id in list(external_id_to_cpg_id.values()):
+            input_vcf_dict[id] = input_dir + '/' + id + '_EH.vcf'
+    elif caller == 'gangstr':
+        for id in list(external_id_to_cpg_id.values()):
+            input_vcf_dict[id] = input_dir + '/' + id + '_gangstr.vcf'
+    else:
+        raise Exception('Invalid caller')
```
MattWellie, line 72:
```suggestion
    for id in list(external_id_to_cpg_id.values()):
        input_vcf_dict[id] = os.path.join(input_dir, f'{id}_{caller}.vcf')
```

This assumes the above change to make the `--caller` argument a strict choice, so you already know to expect one of `gangstr` or `eh`. Change to behaviour here is that `EH` is not capitalised.

str/trtools/merge_str_prep.py
```diff
@@ -0,0 +1,143 @@
...
+                    'vcf.gz': '{root}.vcf.gz',
+                    'vcf.gz.tbi': '{root}.reheader.vcf.gz.tbi',
+                }
+            )
+            bcftools_job.command(
+                f"""
+
+                bcftools sort {vcf_input} | bgzip -c  > {bcftools_job.vcf_sorted['vcf.gz']}
+            
+                tabix -f -p vcf {bcftools_job.vcf_sorted['vcf.gz']} > {bcftools_job.vcf_sorted['vcf.gz.tbi']}
```
MattWellie, line 125:
```suggestion
                tabix -p vcf {bcftools_job.vcf_sorted['vcf.gz']}
```
I don't think tabix `.tbi` outputs can be directed, it's always written to `input_file.tbi`

str/trtools/merge_str_prep.py
```diff
@@ -0,0 +1,143 @@
...
+                }
+            )
+            bcftools_job.command(
+                f"""
+
+                bgzip -c {vcf_input} > {bcftools_job.vcf_sorted['vcf.gz']}
+            
+                bcftools reheader -f {ref.fai} -o {bcftools_job.vcf_sorted['reheader.vcf.gz']} {bcftools_job.vcf_sorted['vcf.gz']} 
+
+                tabix -f -p vcf {bcftools_job.vcf_sorted['reheader.vcf.gz']} > {bcftools_job.vcf_sorted['vcf.gz.tbi']}
```
MattWellie, line 98:
```suggestion
                tabix -f -p vcf {bcftools_job.vcf_sorted['reheader.vcf.gz']}
```

str/trtools/merge_str_runner.py
```diff
@@ -0,0 +1,106 @@
...
+from cpg_utils.hail_batch import remote_tmpdir, output_path
+
+config = get_config()
+
+TRTOOLS_IMAGE = config['images']['trtools']
+
+
+# inputs:
+# caller
+@click.option('--caller', help='gangstr or eh')
```
MattWellie, line 30:
```suggestion
@click.option('--caller', help='gangstr or eh', type=click.Choice(['eh', 'gangstr'], case_sensitive=True))
```

str/trtools/merge_str_runner.py
```diff
@@ -0,0 +1,106 @@
...
+        dataset, list(external_wgs_ids)
+    )
+
+    vcf_input = []
+    if caller == 'eh':
+        for id in list(external_id_to_cpg_id.values()):
+            sample_vcf_file = b.read_input_group(
+                vcf=input_dir + '/' + id + '_eh.reheader.vcf.gz',
+                tbi=input_dir + '/' + id + '_eh.reheader.vcf.gz.tbi',
+            )
```
MattWellie, line 59:
```suggestion
                vcf=os.path.join(input_dir, f'{id}_eh.reheader.vcf.gz'),
                tbi=os.path.join(input_dir, f'{id}_eh.reheader.vcf.gz.tbi'),
            )
```

str/trtools/merge_str_runner.py
```diff
@@ -0,0 +1,106 @@
...
+            vcf_input.append(sample_vcf_file.vcf)
+
+    elif caller == 'gangstr':
+        for id in list(external_id_to_cpg_id.values()):
+            sample_vcf_file = b.read_input_group(
+                vcf=input_dir + '/' + id + '_gangstr.vcf.gz',
+                tbi=input_dir + '/' + id + '_gangstr.vcf.gz.tbi',
```
MattWellie, line 66:
```suggestion
                vcf=os.path.join(input_dir, f'{id}_gangstr.vcf.gz'),
                tbi=os.path.join(input_dir, f'{id}_gangstr.vcf.gz.tbi'),
```

str/trtools/merge_str_runner.py
```diff
@@ -0,0 +1,106 @@
...
+
+    i = 0
+    while i < len(vcf_input):
+        if i != len(vcf_input) - 1:
+            multi_vcf_file_path_string = (
+                multi_vcf_file_path_string + str(vcf_input[i]) + ','
+            )
+        else:
+            multi_vcf_file_path_string = multi_vcf_file_path_string + str(vcf_input[i])
+        i += 1
```
MattWellie, line 81:
```suggestion
    multi_vcf_file_path_string = ','.join(str(vcf_path) for vcf_path in vcf_input)
```

str/trtools/merge_str_runner.py
```diff
@@ -0,0 +1,106 @@
...
+    trtools_job.command(
+        f"""
+     
+    mergeSTR --vcfs {multi_vcf_file_path_string} --out {trtools_job.ofile} --vcftype {caller}
+     
+    """
+    )
+    num_samples = len(vcf_input)
+
+    output_path_vcf = output_path(f'mergeSTR_{num_samples}_samples_{caller}')
```
MattWellie, line 99: should this output file have a VCF/gz file extension?

MattWellie (commented):
Sorry, I haven't gotten to the final 2 (QC/Stat)

str/trtools/compare_str_runner.py
```diff
@@ -0,0 +1,113 @@
...
+    echo "compressing {vcf_input_2}"; 
+    bcftools sort {vcf_input_2} | bgzip -c >{bcftools_job.vcf_2['vcf.gz']};
+    
+    echo "indexing {bcftools_job.vcf_2['vcf.gz']}";
+    tabix -p vcf {bcftools_job.vcf_2['vcf.gz']};
+    """
+    )
+
+    trtools_job = b.new_job(name=f'compareSTR')
+    trtools_job.image(TRTOOLS_IMAGE)
```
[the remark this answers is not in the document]
MattWellie: This could apply to some other scripts as well, but I think in the others I looked at you were preparing inputs/outputs and running one job, which doesn't require workflow dependencies

str/trtools/merge_str_runner.py
hopedisastro: This saves so many lines!

str/trtools/merge_str_prep.py
hopedisastro: added suggestion to next commit

str/trtools/merge_str_prep.py
hopedisastro: added to next commit

str/trtools/merge_str_prep.py
hopedisastro: It seems to work with `cpg-tob-wgs-test` data?

str/trtools/merge_str_runner.py
hopedisastro: added to commit

str/trtools/merge_str_runner.py
hopedisastro: I think it's ok because I declared the resource group:

str/trtools/merge_str_prep.py
MattWellie: It'll write it to that location, but tabix as a tool doesn't let you direct output to a specific file - it always just creates a file called (input_filename) + 'tbi'. 

This change isn't going to change the created file, but it's just good to point out that when you specify the filename here it isn't doing anything.

str/trtools/qc_str_runner.py
```diff
@@ -0,0 +1,122 @@
...
+    help='Exclude points computed using fewer than this many genotypes. This option is meant to avoid plotting outlier points driven by bins with small numbers of TRs with that reference length. Default=100',
+)
+@click.option(
+    '--refbias-xrange-min',
+    help='Exclude points corresponding to TRs with reference length less than this value. Default = 0',
+)
+@click.option(
+    '--refbias-xrange-max',
+    help='Exclude points corresponding to TRs with reference length greater than this value. Default = 10000',
+)
```
MattWellie, line 54:
```suggestion
@click.option(
    '--refbias-binsize',
    help=' Sets the binsize (in bp) used to bin x-axis values, which give the reference TR length', default=5, type=int
)
@click.option(
    '--refbias-metric',
    help='Determines which metric to use to summarize the reference bias in each bin.', type=click.Choice(['mean', 'median'], case_sensitive=True), default='mean'
)
@click.option(
    '--refbias-mingts',
    help='Exclude points computed using fewer than this many genotypes. This option is meant to avoid plotting outlier points driven by bins with small numbers of TRs with that reference length.', type=int, default=100
)
@click.option(
    '--refbias-xrange-min',
    help='Exclude points corresponding to TRs with reference length less than this value.', type=int, default=0
)
@click.option(
    '--refbias-xrange-max',
    help='Exclude points corresponding to TRs with reference length greater than this value.', type=int, default=10000
)
```

Instead of just describing the options you can include default values, meaning that you won't need to use every flag when you're running the script. This also has the benefit of being self-describing (type, default)

str/trtools/stat_str_runner.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
+# pylint: disable=duplicate-code
+"""
+This script runs statSTR() from TRTools package on a single/merged STR vcf file and outputs various statistics
+
+For example: 
+analysis-runner --access-level test --dataset tob-wgs --description 'tester' --output-dir 'tester' statSTR_runner.py --caller=eh --file-path=gs://cpg-tob-wgs-test/hoptan-str/mergeSTR/mergeSTR_2_samples_gangstr.vcf
+
+Required packages: sample-metadata, hail, click, os
```
MattWellie, line 9: I think each of your scripts has this in comments - it's better to formalise this by having a requirements.txt file in the project root, ideally with versions of the relevant libraries.

sample-metadata>=X.Y.z
hail>=0.2.105
...

they can then be installed in bulk with

```
pip install -r requirements.txt
```

MattWellie (approved):

PR #61, opened 2023-02-02
SPEED

MattWellie:
Removes a few `non-pythonic` inclusions:

* instead of iterating over objects by index, iterate over the object exactly, e.g. instead of a `while k <= len(object); row = object[k]`, just do `for row in object` - much easier to read, mostly foolproof (will just stop looping once there are no more rows), and will work on just about every iterable object type (lists, tuples, dicts, file handlers...)
* Don't store more than you have to - here we have the first 4 lines that we need repeatedly (metadata), but each other line is handled atomically. That means that we don't need to hold it all in memory. Reading all the contents into memory and iterating by index is MUCH slower than reading a file one line at a time, but for this purpose accomplishes the same result.
* Think about your output, and then the easiest way to reach it. In the original you're constructing a pandas dataframe (a 'heavy' object, with a lot of features), then appending to it a million times. This potentially means a million regenerations of an index, and provisioning new memory space for the growing object. Instead of that, just write directly to an output file line by line, with a negligible memory footprint.

This revised script generates ~78MB of output, 1.3M lines, in 0.7 seconds

### DIFF
--- a/str/sensitivity-analysis/capillary_genotypes.py
+++ b/str/sensitivity-analysis/capillary_genotypes.py
@@ -1,57 +1,74 @@
 import warnings
+
+
 warnings.simplefilter(action='ignore', category=FutureWarning)
-import pandas as pd 
-
-file = open("combinedmicrosats_627loci_1048indivs_numRpts.stru.txt")
-file = file.readlines()
-
-loci = file[0].split()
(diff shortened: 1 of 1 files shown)


MattWellie:
closing this... over a year old, can't remember why it exists

PR #62, opened 2023-02-19
Create data_frame_creator.py

hopedisastro:
This PR adds a data_frame_creator.py and updates the existing EH and GangSTR runners so that `hgdp` samples are run using the unmasked fasta.

### DIFF
--- a/str/data_frame_creator.py
+++ b/str/data_frame_creator.py
@@ -14,6 +14,7 @@ import hailtop.batch as hb
 from cpg_utils import to_path
 from google.cloud import storage
 import os
+from cloudpathlib import GSPath
 
 config = get_config()
 
@@ -48,8 +49,8 @@ def eh_csv_writer(input_dir):
 
     for file in files:
         if isinstance(file, GSPath):
-            file.copy('localfile.vcf')
-            file = 'localfile.vcf'
+            file.copy(file.name)
+            file = file.name
         sample_id = str(VCFReader(file).samples[0])
         for variant in VCFReader(file):
             chr = str(variant.CHROM)
@@ -123,8 +124,8 @@ def gangstr_tsv_writer(input_dir):
     )
     for file in files:
         if isinstance(file, GSPath):
-            file.copy('localfile.vcf')
-            file = 'localfile.vcf'
+            file.copy(file.name)
+            file = file.name
         sample_id = str(VCFReader(file).samples[0])
         for variant in VCFReader(file):
             chr = str(variant.CHROM)
--- a/str/data_frame_creator.py
+++ b/str/data_frame_creator.py
@@ -149,18 +149,18 @@ def gangstr_tsv_writer(input_dir):
             g_dp = str(variant.format('DP')[0][0])
             g_q = str(variant.format('Q')[0][0])
             g_repcn = (
-                f'{variant.format('REPCN')[0][0]}, {variant.format('REPCN')[0][1]}'
+                f'{variant.format("REPCN")[0][0]}, {variant.format("REPCN")[0][1]}'
             )
             g_repci = str(variant.format('REPCI')[0])
             g_rc = str(variant.format('RC')[0])
             g_enclreads = str(variant.format('ENCLREADS')[0])
             g_flnkreads = str(variant.format('FLNKREADS')[0])
             g_ml = str(variant.format('ML')[0][0])
-            g_ins = f'{variant.format('INS')[0][0]},{variant.format('INS')[0][1]}'
+            g_ins = f'{variant.format("INS")[0][0]},{variant.format("INS")[0][1]}'
             g_stderr = (
-                f'{variant.format('STDERR')[0][0]},{variant.format('STDERR')[0][1]}'
+                f'{variant.format("STDERR")[0][0]},{variant.format("STDERR")[0][1]}'
             )
-            g_qexp = f'{variant.format('QEXP')[0][0]},{variant.format('QEXP')[0][1]},{variant.format('QEXP')[0][2]}'
+            g_qexp = f'{variant.format("QEXP")[0][0]},{variant.format("QEXP")[0][1]},{variant.format("QEXP")[0][2]}'
             g_allele_1 = str(g_repcn.split(',')[0].strip())
             g_allele_2 = str(g_repcn.split(',')[1].strip())
             csv = csv + (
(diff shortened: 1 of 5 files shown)


str/data_frame_creator.py
```diff
@@ -0,0 +1,239 @@
...
+config = get_config()
+
+
+def eh_csv_writer(input_dir):
+    """Creates a CSV file containing dataframe of merged EH VCFs"""
+    bucket_name, *components = input_dir[5:].split('/')
+    client = storage.Client()
+    bucket = client.bucket(bucket_name)
+    blobs = client.list_blobs(bucket_name, prefix='/'.join(components))
+    files = {f'gs://{bucket_name}/{blob.name}' for blob in blobs}
```
MattWellie, line 23:
```suggestion
    files = to_path(input_dir).glob('*.vcf')
```

this is using `cloudpathlib.AnyPath/CloudPath` or `cpg_utils.to_path`. This also removes the check you have a few lines later where you only consider VCF files from the list of returned blobs

You can also open and read these results directly

```
for file in files:
    with file.open() as f:
         ...
```

There are reasons to avoid using cloudpathlib methods like this, but for your purpose here we are trying to retrieve and read all contents in the folder, so I don't think the downsides are super relevant

str/data_frame_creator.py
```diff
@@ -0,0 +1,239 @@
...
+        + '\n'
+    )
+
+    for file in files:
+        if file.endswith('.vcf'):
+            blob = bucket.blob(file[6 + len(bucket_name) :])
+            with blob.open('r') as f:
+                array = f.readlines()
+                for line in array:
+                    if line.startswith('##'):
```
MattWellie, line 55: This whole section will work, but please be aware of VCF parsing libraries that will streamline this whole process massively, e.g. [pyvcf](https://pyvcf.readthedocs.io/en/latest/INTRO.html) or CyVCF2

str/data_frame_creator.py
```diff
@@ -0,0 +1,239 @@
...
+                    )
+    return csv
+
+
+def gangstr_csv_writer(input_dir):
+    """Creates a TSV file containing dataframe of merged GangSTR VCFs"""
+    bucket_name, *components = input_dir[5:].split('/')
+    client = storage.Client()
+    bucket = client.bucket(bucket_name)
+    blobs = client.list_blobs(bucket_name, prefix='/'.join(components))
```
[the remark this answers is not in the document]
MattWellie: This wouldn't be required if the cloudpathlib change above is used, but is probably good to know anyway

str/data_frame_creator.py
```diff
@@ -0,0 +1,239 @@
...
+        remote_tmpdir=remote_tmpdir(),
+    )
+    b = hb.Batch(
+        backend=backend, default_python_image=config['workflow']['driver_image']
+    )
+    j = b.new_python_job(name='EH dataframe writer')
+    g = b.new_python_job(name='GangSTR dataframe writer')
+
+    eh_csv = j.call(eh_csv_writer, input_dir_eh)
+    gangstr_csv = g.call(gangstr_csv_writer, input_dir_gangstr)
```
MattWellie, line 230: As above - is there reason for one to be a TSV and one a CSV? If so, the method names should indicate which is which

str/data_frame_creator.py
```diff
@@ -0,0 +1,239 @@
+#!/usr/bin/env python3
+# pylint: disable=import-error, too-many-locals, broad-exception-raised
+"""
+This script merges all the VCFs from one STR caller into a .CSV (ExpansionHunter) or .TSV (GangSTR) format that can be read into R.
```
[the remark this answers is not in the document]
hopedisastro: One of GangSTR's attributes is in a tuple-style format eg (X,Y,Z) so if I store it as a CSV, it gets separated into multiple columns.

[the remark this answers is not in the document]
hopedisastro: Took me a few tries to realise why the output was wonky :(

str/data_frame_creator.py
hopedisastro: This is neat. I think I was hesitating to use `cloudpathlib` because it's harder to see what's happening underneath the hood, and was OK with the chunkier method as long as it was working. Will use `to_path` from now on.

str/data_frame_creator.py
hopedisastro: Haha, I didn't even realise this

str/data_frame_creator.py
```diff
@@ -0,0 +1,222 @@
...
+            ]
+        )
+        + '\n'
+    )
+    for file in files:
+        if isinstance(file, GSPath):
+            file.copy(file.name)
+            file = file.name
+        sample_id = str(VCFReader(file).samples[0])
+        for variant in VCFReader(file):
```
MattWellie, line 128:
```suggestion
        reader = VCFReader(file)
        sample_id = str(reader.samples[0])
        for variant in reader:
```

Not a big performance hit, but when you create a VCFReader it absorbs all the Header data, Samples, etc. Here you're creating one populated object, throwing it away, then repeating with the second object.

str/data_frame_creator.py
```diff
@@ -0,0 +1,222 @@
...
+            chr = str(variant.CHROM)
+            start = str(variant.POS)
+            g_gt = str(variant.format('GT')[0])
+            if g_gt == '':  # ie variant is not called
```
MattWellie, line 132: you're populating all these attributes, then throwing them away by adding `continue` at the end (which will skip to the next variant)

Do you mean to add these to the output as a row?

str/data_frame_creator.py
```diff
@@ -0,0 +1,222 @@
...
+        )
+        + '\n'
+    )
+
+    for file in files:
+        if isinstance(file, GSPath):
+            file.copy(file.name)
+            file = file.name
+        sample_id = str(VCFReader(file).samples[0])
+        for variant in VCFReader(file):
```
MattWellie, line 53:
```suggestion
        reader = VCFReader(file)
        sample_id = str(reader.samples[0])
        for variant in reader:
```

MattWellie (commented):
Couple of things to check

str/data_frame_creator.py
hopedisastro: Yes

hopedisastro:
I keep forgetting to add more descriptive commit messages. Basically added the `reader` variable to both writer functions and ensured that the not-called variants still get written out to the CSV.

MattWellie (approved):

PR #63, opened 2023-02-19
Hipstr runner

hopedisastro:
There are 2 linting checks that have failed and are found in files that are not included in this PR (`merge-str-runner` and `qc-str-runner`. Those checks have been addressed in another PR currently under review.


str/str_iterative_hipstr_runner.py
```diff
@@ -0,0 +1,143 @@
...
+            base=ref_fasta,
+            fai=ref_fasta + '.fai',
+            dict=ref_fasta.replace('.fasta', '').replace('.fna', '').replace('.fa', '')
+            + '.dict',
+        )
+    )
+
+    hipstr_job.command(
+        f"""
+    HipSTR --bams {cramfuse_path} --fasta {ref.base} --regions {hipstr_regions} --str-vcf {hipstr_job.hipstr_output['vcf.gz']} --viz-out {hipstr_job.hipstr_output['viz.gz']} --log {hipstr_job.hipstr_output['log.txt']} --output-filters
```
illusional, line 110: Nitpick, but I think it's cleaner to keep these on separate lines, and makes looking at the command in Hail Batch easier:

```suggestion
    HipSTR --bams {cramfuse_path} \\
        --fasta {ref.base} \\
        --regions {hipstr_regions} \\
        --str-vcf {hipstr_job.hipstr_output['vcf.gz']} \\
        --viz-out {hipstr_job.hipstr_output['viz.gz']} \\
        --log {hipstr_job.hipstr_output['log.txt']} \\
        --output-filters
```

str/str_iterative_hipstr_runner.py
```diff
@@ -0,0 +1,143 @@
...
+    b.write_output(hipstr_job.hipstr_output['viz.gz'], hipstr_output_path_viz)
+
+    hipstr_output_path_log = output_path(f'{output_file_name}_hipstr.log.txt')
+    b.write_output(hipstr_job.hipstr_output['log.txt'], hipstr_output_path_log)
+
+    #Unzip .VCF.GZ output into .VCF
+    samtools_job = b.new_job(name='Unzip VCF')
+    samtools_job.image(SAMTOOLS_IMAGE)
+    samtools_job.storage('20G')
+    samtools_job.cpu(8)
```
illusional, line 127: DDo you need to write the uncompressed VCF (increases storage cost)? If tools in the future need the uncompressed, I would imagine it's better to bgzip on the fly.

illusional (changes requested):

str/str_iterative_hipstr_runner.py
```diff
@@ -0,0 +1,143 @@
...
+@click.command()
+def main(
+    variant_catalog, dataset, external_wgs_ids, output_file_name
+):  # pylint: disable=missing-function-docstring
+    b = get_batch()
+    #Create HipSTR job 
+    hipstr_job = b.new_job(name=f'HipSTR running')
+    hipstr_job.image(HIPSTR_IMAGE)
+    hipstr_job.storage('375G')
+    hipstr_job.cpu(16)
```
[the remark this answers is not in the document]
hopedisastro: This was the memory profile for 91 samples: https://batch.hail.populationgenomics.org.au/batches/421353/jobs/1
I'm planning to run it on 150..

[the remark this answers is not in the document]
hopedisastro: How do I calculate how much memory and CPUs I need?

hopedisastro:
I also got a new linting error after making the changes..but not sure how to fix it:

str/str_iterative_hipstr_runner.py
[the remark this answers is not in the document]
illusional: That batch looks like it's only using 1 CPU (single threaded), but it's using a bit of memory. Hard to really gauge how much memory it will use, it might be opportunistically allocating memory (eg: it will happily use less memory).

You might consider using the `highmem` machine, then either 4 / 8 cores (which gives 26GB / 52GB RAM).

str/str_iterative_hipstr_runner.py
```diff
@@ -0,0 +1,137 @@
...
+        external_wgs_sids_without_crams = ', '.join(
+            cpg_id_to_external_id[sid] for sid in cpg_sids_without_crams
+        )
+        logging.warning(
+            f'There were some samples without CRAMs: {external_wgs_sids_without_crams}'
+        )
+    # Create string containing paths based on /cramfuse
+    cramfuse_path = ''
+    for cram_obj in crams_path:
+        cpg_sample_id = os.path.basename(cram_obj['output']).split('.')[0]
```
illusional, line 68: Can you use the path instead of trying to extract the filename. 
Because you're mounting the main bucket directly, then you could just strip the bucket name and use the path, eg:

```python
# removes the gs:// then, grab everything after the bucket name
suffix = cram_obj['output'].removeprefix('gs://').split('/', maxsplit=1)[1]
cramfuse_path = f'/cramfuse/{suffix}'
```

str/str_iterative_hipstr_runner.py
```diff
@@ -0,0 +1,137 @@
...
+    cpg_sids_without_crams = set(cpg_id_to_external_id.keys()) - cpg_sids_with_crams
+    if cpg_sids_without_crams:
+        external_wgs_sids_without_crams = ', '.join(
+            cpg_id_to_external_id[sid] for sid in cpg_sids_without_crams
+        )
+        logging.warning(
+            f'There were some samples without CRAMs: {external_wgs_sids_without_crams}'
+        )
+    # Create string containing paths based on /cramfuse
+    cramfuse_path = ''
```
illusional, line 66: Better to store a list of filenames here, then join them at the end (`','.join(cram_paths)`)

illusional (changes requested):

str/str_iterative_hipstr_runner.py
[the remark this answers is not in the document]
hopedisastro: How do I set the job to `highmem` exactly? I tried looking for examples in the popgen repo and Hail Batch documentation but couldn't find an example for this context.

str/str_iterative_hipstr_runner.py
```diff
@@ -0,0 +1,142 @@
...
+            status=AnalysisStatus('completed'),
+            meta={'sequence_type': 'genome', 'source': 'nagim'},
+        )
+    else: 
+        analysis_query_model = AnalysisQueryModel(
+            sample_ids=list(external_id_to_cpg_id.values()),
+            projects=[dataset],
+            type=AnalysisType('cram'),
+            status=AnalysisStatus('completed'),
+            meta={},
```
illusional, line 52: I think you still need sequencing_type: genome here

str/str_iterative_hipstr_runner.py
```diff
@@ -0,0 +1,142 @@
...
+    external_id_to_cpg_id: dict[str, str] = SampleApi().get_sample_id_map_by_external(
+        dataset, list(external_wgs_ids)
+    )
+    if dataset == 'tob-wgs':
+        analysis_query_model = AnalysisQueryModel(
+            sample_ids=list(external_id_to_cpg_id.values()),
+            projects=[dataset],
+            type=AnalysisType('cram'),
+            status=AnalysisStatus('completed'),
+            meta={'sequence_type': 'genome', 'source': 'nagim'},
```
illusional, line 44: Context: https://centrepopgen.slack.com/archives/C020PDNH3N0/p1676953947345959
```suggestion
            meta={'sequencing_type': 'genome', 'source': 'nagim'},
```

str/str_iterative_hipstr_runner.py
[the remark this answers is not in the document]
illusional: This should do it :)
```suggestion
    hipstr_job.cpu(4)
    j.memory('highmem')
```

str/str_iterative_hipstr_runner.py
```diff
@@ -0,0 +1,142 @@
...
+    hipstr_job.cpu(16)
+
+    hipstr_job.declare_resource_group(
+        hipstr_output={
+            'vcf.gz': '{root}.vcf.gz',
+            'viz.gz': '{root}.viz.gz',
+            'log.txt': '{root}.log.txt',
+        }
+    )
+    hipstr_job.cloudfuse(f'cpg-{dataset}-main', '/cramfuse')
```
illusional, line 99: @lgruen, is there any issue with binding the whole main disk on? Does it do anything if you don't access any files, or does it cause a listing if you only access specific files?

If so, you could simplify using `os.path.commonprefix`

illusional (commented):
Looks pretty good, a few minor touch-ups + waiting on a comment from Leo to approve.

str/str_iterative_hipstr_runner.py
lgruen: Mounting the whole bucket should be okay, i.e. it shouldn't cause a full listing. The semantics with GCSFuse can be a bit tricky, see e.g. https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/docs/semantics.md#implicit-directories.

str/str_iterative_hipstr_runner.py
hopedisastro: with and without sequencing_type: genome gives the same output on a test case I just tried out

illusional (approved):

PR #66, opened 2023-02-23
Hipstr cv writer

hopedisastro:
adds hipstr_cv_writer method to the data_frame_creator.py

### DIFF
--- a/str/data_frame_creator.py
+++ b/str/data_frame_creator.py
@@ -193,7 +193,7 @@ def gangstr_tsv_writer(input_dir):
 
 def hipstr_csv_writer(input_dir):
     """Creates a CSV file containing dataframe of merged HipSTR VCFs"""
-    file = to_path(input_dir).glob('*.vcf.gz')  # HipSTR output is a merged vcf.gz file
+    files = to_path(input_dir).glob('*.vcf.gz')  # HipSTR output is a merged vcf.gz file
     csv = (
         ','.join(
             [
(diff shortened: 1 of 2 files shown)


hopedisastro:
Had to use a `while` loop (not by choice, but out of necessity I think??)

MattWellie (approved):

str/data_frame_creator.py
```diff
@@ -191,14 +191,89 @@ def gangstr_tsv_writer(input_dir):
...
+            samples_dict = {}
+            i = 0
+            while i < len(reader.samples):
+                sample = reader.samples[i]
+                gb = variant.format('GB')[i]
+                q = variant.format('Q')[i][0]
+                samples_dict[sample] = (gb, q)
+                i += 1
+            for sample in samples_dict:
+                gb, q = samples_dict[sample]
```
MattWellie, line 236:
```suggestion
            for sample, gb, q in zip(
                    reader.samples, 
                    variant.format('GB'), 
                    variant.format('Q')
            ):
                q = q[0]
```

This should be a viable alternative - using zip you can align multiple objects of the same length and iterate them all as trios.

Another alternative would be to use `enumerate` - a really nice built-in function that lets you iterate over something whilst keeping the index:

```
my_list = [1, 2, 3, 4]
for i, value in enumerate(my_list):
    print(i, value)

> 0, 1
> 1, 2
...

Comment (MattWellie) [inline str/data_frame_creator.py:236]: I think the linter's original complaint is here. Instead of using 

```
for sample in dict:
    item = dict[sample]...
```

python has `items`, `values`, and `keys` as dictionary operations, e.g.

```
for sample, content in dict.items():
    gq = content[gq]
```

PR #68, opened 2023-02-26
Reviewer runner - visualises EH read pile ups 

hopedisastro:


str/reviewer_runner.py
```diff
@@ -0,0 +1,101 @@
...
+"""
+import os
+import click
+
+
+from cpg_utils.config import get_config
+from cpg_utils.hail_batch import output_path
+from cpg_workflows.batch import get_batch
+
+REF_FASTA = 'gs://cpg-common-main/references/hg38/v0/Homo_sapiens_assembly38.fasta'
```
illusional, line 15: Can you pull this from the config?

str/reviewer_runner.py
```diff
@@ -0,0 +1,101 @@
...
+config = get_config()
+
+SAMTOOLS_IMAGE = config['images']['samtools']
+REVIEWER_IMAGE = config['images']['reviewer']
+
+
+@click.option('--locus', help='Locus identifier as per catalog')
+@click.option('--catalog', help='GCP path to catalog')
+@click.option('--input-dir', help='GCP path to input-dir, includes gs://')
+@click.argument('cpg-wgs-ids', nargs=-1)
```
illusional, line 26: Would recommend a slight nomenclature change 
```suggestion
@click.argument('cpg-sample-ids', nargs=-1)
```

str/reviewer_runner.py
```diff
@@ -0,0 +1,101 @@
...
+        samtools_job.declare_resource_group(
+            bam={
+                'sorted.bam': '{root}.sorted.bam',
+                'sorted.bam.bai': '{root}.sorted.bam.bai',
+            }
+        )
+
+        samtools_job.command(
+            f"""
+
```
illusional, line 62:
```suggestion
```

str/reviewer_runner.py
```diff
@@ -0,0 +1,101 @@
...
+        bam_input = b.read_input(f'{input_dir}/{cpg_id}_eh.realigned.bam')
+
+        # read in VCF from EH output
+        vcf_input = b.read_input(f'{input_dir}/{cpg_id}_eh.vcf')
+
+        # BAM file must be sorted and indexed prior to inputting into REViewer
+        samtools_job = b.new_job(name=f'Sorting and indexing {cpg_id}')
+        samtools_job.image(SAMTOOLS_IMAGE)
+        samtools_job.storage('20G')
+        samtools_job.cpu(8)
```
illusional, line 52: Generally worth optimising resources here, samtools sorting can be multithreaded, but you're not using the threaded option below. Can you dynamically size the storage based on the input size of the cram (and flow it through to the bottom)?

str/reviewer_runner.py
```diff
@@ -0,0 +1,101 @@
...
+            ofile={
+                'svg': '{root}.' + f'{locus}.svg',
+                'metrics.tsv': '{root}.metrics.tsv',
+                'phasing.tsv': '{root}.phasing.tsv',
+            }
+        )
+
+        reviewer_job.command(
+            f"""
+            ./REViewer-v0.2.7-linux_x86_64 --reads {samtools_job.bam['sorted.bam']} --vcf {vcf_input} --reference {ref.base} --catalog {ref.catalog} --out {reviewer_job.ofile} --locus {locus}
```
illusional, line 89: IMO (nitpick): it would be worth formatting these on separate lines to make the command easier to read later

str/reviewer_runner.py
```diff
@@ -0,0 +1,101 @@
...
+
+        """
+        )
+
+        # REViewer job
+        reviewer_job = b.new_job(name=f'Visualising {locus} for {cpg_id}')
+        reviewer_job.image(REVIEWER_IMAGE)
+        reviewer_job.depends_on(samtools_job)
+        reviewer_job.storage('20G')
+        reviewer_job.cpu(8)
```
illusional, line 77: Similar comment about appropriate resources, do you need 8 cores, what sort of memory requirements are useful?

illusional (changes requested):

str/reviewer_runner.py
hopedisastro: The masked fasta can be pulled with: ref_fasta = reference_path('broad/ref_fasta')

str/reviewer_runner.py
hopedisastro: How do I dynamically size the storage based on the input size of the cram?

str/reviewer_runner.py
hopedisastro: Using the default options when initialising the batch job seems to be okay

str/reviewer_runner.py
illusional: Ah okay, I won't push it for this PR, but could you add an "ref_fasta_unmasked" to https://github.com/populationgenomics/references when you get a moment?

str/reviewer_runner.py
illusional: Yeah that should be sweet then

str/reviewer_runner.py
```diff
@@ -0,0 +1,97 @@
...
+    b = get_batch()
+    ref = b.read_input_group(
+        **dict(
+            base=REF_FASTA,
+            catalog=catalog,
+            fai=REF_FASTA + '.fai',
+            dict=REF_FASTA.replace('.fasta', '').replace('.fna', '').replace('.fa', '')
+            + '.dict',
+        )
+    )
```
illusional, line 40: You can just provide these options straight away, without the `**dict(**params)`, eg:

```suggestion
    ref = b.read_input_group(
        base=REF_FASTA,
        catalog=catalog,
        fai=REF_FASTA + '.fai',
        dict=REF_FASTA.replace('.fasta', '').replace('.fna', '').replace('.fa', '')
            + '.dict',
    )
```

str/reviewer_runner.py
```diff
@@ -0,0 +1,97 @@
...
+    for cpg_id in cpg_sample_ids:
+        # read in bam file from EH output
+        bam_input = b.read_input(f'{input_dir}/{cpg_id}_eh.realigned_bam')
```
illusional, line 43:
```suggestion
        bam_input = b.read_input(os.path.join(input_dir, f'{cpg_id}_eh.realigned_bam'))
```

str/reviewer_runner.py
```diff
@@ -0,0 +1,97 @@
...
+
+        # read in VCF from EH output
+        vcf_input = b.read_input(f'{input_dir}/{cpg_id}_eh.vcf')
```
illusional, line 46:
```suggestion
        vcf_input = b.read_input(os.path.join(input_dir, f'{cpg_id}_eh.vcf'))
```

illusional (changes requested):
Looking good to me, I just want to see the final resources for samtools (+ using the `-@` parameter for threads) before approving.

str/reviewer_runner.py
hopedisastro: yes

str/reviewer_runner.py
```diff
@@ -0,0 +1,106 @@
...
+        reviewer_job.command(
+            f"""
+            ./REViewer-v0.2.7-linux_x86_64 --reads {samtools_job.bam['sorted.bam']} \\
+            --vcf {vcf_input} --reference {ref.base} --catalog {ref.catalog} \\
+            --out {reviewer_job.ofile} --locus {locus}
+            """
+        )
+    # output writing
+    reviewer_output_path = output_path(f'{locus}/{cpg_id}_{locus}_reviewer', 'analysis')
+    b.write_output(reviewer_job.ofile, reviewer_output_path)
```
illusional, line 100: These should be inside the loop, otherwise you're only writing one output from all jobs. This is what the `undefined-loop-variable` was warning you against.

illusional (approved):
Nice work @hopedisastro!
Once the last review is addressed, happy for this to be merged, definitely make sure the writing inside the loop is addressed.

str/reviewer_runner.py
```diff
@@ -0,0 +1,106 @@
...
+        )
+
+        # REViewer job
+        reviewer_job = b.new_job(name=f'Visualising {locus} for {cpg_id}')
+        reviewer_job.image(REVIEWER_IMAGE)
+        reviewer_job.depends_on(samtools_job)
+
+        reviewer_job.declare_resource_group(
+            ofile={
+                'svg': '{root}.' + f'{locus}.svg',
```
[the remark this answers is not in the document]
hopedisastro: I tried this before, but then it thinks `root` is a variable that has been defined, and so gives me an error.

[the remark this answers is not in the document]
illusional: Ohhh, sorry I suggested this wrong, you can escape with a double `{{`, eg:

```suggestion
f'{{root}}.{locus}.svg',
```

hopedisastro:
Thanks for your help @illusional :)

PR #72, opened 2023-04-10
ExpansionHunter runner for TOB cohort

hopedisastro:
This PR modifies generic runner to run EH specifically for TOB cohort. 

A modified script is needed because EH requires karyotypic sex which is not available via metamist, but is found in gs://cpg-tob-wgs-main-analysis/joint-calling/v7/meta.tsv

There are two files in this PR. The simple R script wrangles the `meta.tsv` file to produce a clean CSV file containing only CPG sample id, karyotypic sex, and TOB id, which was uploaded here: gs://cpg-tob-wgs-test/hoptan-str/karyotype_sex_mapping.csv

The sample and sex mapping file will be used as input for the TOB-specific EH runner (the second script in this PR).


hopedisastro:
linting check keeps failing for `AnyPath` because it says that it doesn't have an associated `open()` method, but this command works in tester scripts

PR #74, opened 2023-04-14
Increase `storage` for one sample with assoc big cram

hopedisastro:


str/tob-wgs/tob_eh_runner.py
```diff
@@ -108,7 +108,10 @@ def main(
                 eh_job.depends_on(jobs[-max_parallel_jobs])
             jobs.append(eh_job)
             eh_job.image(EH_IMAGE)
-            eh_job.storage('50G')
+            if cpg_id == 'CPG5454': #46G WGS cram, requiring extra disk space
+                eh_job.storage('70G')
+            else:
+                eh_job.storage('50G')
```
lgruen, line 114: Such special cases are brittle. It's better to either use a value that's large enough for all files or instead adjust the limit based on the input file size.

```suggestion
            eh_job.storage('70G')
```

lgruen (approved):

PR #75, opened 2023-04-19
Cloudfuse added

hopedisastro:
When mounting the cloud bucket,  are all the other files within the bucket visible? 

In order for this to work, mergeSTR needs to localise the associated index `.tbi` file for each sample. This is not an input parameter for the function. The `tbi` file just has to be in the same directory as the `vcf.gz` file which does get inputted as a parameter. 

Without cloudfuse, I had to use `b.read input group` and explicitly read in both the `tbi` and `vcf.gz` files for the script to work. Just want to check if the case is the same when using `cloudfuse`?

### DIFF
--- a/str/trtools/merge_str_runner.py
+++ b/str/trtools/merge_str_runner.py
@@ -50,11 +50,11 @@ def main(
     trtools_job.cloudfuse(f'cpg-{dataset}-main', '/vcffuse')
 
     vcffuse_path = []
-    num_samples = len(vcffuse_path)
     for id in list(external_id_to_cpg_id.values()):
         vcf = os.path.join(input_dir, f'{id}_eh.reheader.vcf.gz')
         suffix = vcf.removeprefix('gs://').split('/', maxsplit=1)[1]
         vcffuse_path.append(f'/vcffuse/{suffix}')
+    num_samples = len(vcffuse_path)
     vcffuse_path = ','.join(vcffuse_path)  # string format for input into mergeSTR
 
     output_path_vcf = output_path(f'mergeSTR_{num_samples}_samples_eh.vcf')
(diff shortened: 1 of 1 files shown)


hopedisastro:
linting keeps failing due to black reformatting, but I've already run and committed the changes made by `black`?

str/trtools/merge_str_runner.py
```diff
@@ -49,43 +44,31 @@ def main(
...
+        vcffuse_path.append(f'/vcffuse/{suffix}')
+    vcffuse_path = ','.join(vcffuse_path)  # string format for input into mergeSTR
 
     trtools_job.declare_resource_group(ofile={'vcf': '{root}.vcf'})
 
     trtools_job.command(
         f"""
      
-    mergeSTR --vcfs {multi_vcf_file_path_string} --out {trtools_job.ofile} --vcftype {caller}
+    mergeSTR --vcfs {vcffuse_path} --out {trtools_job.ofile} --vcftype eh
```
lgruen, line 65: Don't you want to write to the `cloudfuse` bucket here directly instead of using `ofile`?

str/trtools/merge_str_runner.py
```diff
@@ -49,43 +44,31 @@ def main(
...
-                tbi=os.path.join(input_dir, f'{id}_gangstr.vcf.gz.tbi'),
-            )
-            vcf_input.append(sample_vcf_file.vcf)
-    else:
-        raise ValueError('Invalid caller')
-    multi_vcf_file_path_string = ','.join(str(vcf_path) for vcf_path in vcf_input)
-
     trtools_job = b.new_job(name='mergeSTR')
     trtools_job.image(TRTOOLS_IMAGE)
     trtools_job.storage('20G')
```
lgruen, line 49: Do you still need this storage?

lgruen (changes requested):
> When mounting the cloud bucket, are all the other files within the bucket visible?

Yes, everything will be visible.

hopedisastro:
is there any way to make it run faster, or do we expect that using `cloudfuse` should already help to speed up the process? As far as I know, multi-threading is not supported by mergeSTR.

lgruen (approved):

PR #76, opened 2023-04-28
Include compression and indexing of output

MattWellie:
The current approach writes a naked VCF
This change writes the VCF to a temporary location, then bgzips and indexes into fused cloud storage

This could be tweaked a bit further to make sure all the outputs are stored - probably a good idea

### DIFF
--- a/str/trtools/merge_str_runner.py
+++ b/str/trtools/merge_str_runner.py
@@ -27,6 +27,7 @@ TRTOOLS_IMAGE = config['images']['trtools']
 
 # inputs:
 
+
 # dataset
 @click.option('--dataset', help='dataset eg tob-wgs')
 # input directory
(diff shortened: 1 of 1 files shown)


hopedisastro:
I don't think bgzip and tabix are part of trtools? Wouldn't we need a separate job with `bcftools`?

MattWellie:
@hopedisastro I didn't think so either! I pulled down the image and had a look and both of them are installed under micromamba. I'm not 10000% sure, but I think this tracks as:

- micromamba installs TRTools
- TRTools has a dependency on PySAM
- PySAM is a wrapper around the HTSlib API (same as samtools, bcftools, and tabix... though BCFTools is not installed)

Screenshot 2023-04-28 at 12 25 37 pm

hopedisastro (approved):

PR #80, opened 2023-05-17
more storage space needed

hopedisastro:


MattWellie:
From this batch log (https://batch.hail.populationgenomics.org.au/batches/424231/jobs/1) the consumption is in memory, not storage. I'm not really sure why that's the case... the copy should make a local version of the file, not just read it into RAM...

MattWellie:
A change I'd suggest is to the copy... Instead of storing all files, we only need to store the single file we're currently reading

```
local_file = 'local.vcf'
for file in files:
        if not isinstance(file, GSPath):
            logging.warning(f'There is a file path that is not a GSPath: {file}')
            continue
        
        file.copy(local_file)
        reader = VCFReader(local_file)
```

MattWellie (approved):
TBH I don't really know why this is failing in the way that it's failing...

PR #81, opened 2023-07-05
Eh graphquery modification

hopedisastro:
Edit script so that it is compatible with new GraphQuery functionality of metamist

### DIFF
--- a/str/runners/str_iterative_eh_runner.py
+++ b/str/runners/str_iterative_eh_runner.py
@@ -91,12 +91,11 @@ def main(
     eh_regions = b.read_input(variant_catalog)
 
     # Iterate over each sample to call Expansion Hunter
-    for cram_obj in crams_path:
+    for cpg_sample_id, cram_obj in crams_by_id.items():
         # Making sure Hail Batch would localize both CRAM and the correponding CRAI index
         crams = b.read_input_group(
             **{'cram': cram_obj['output'], 'cram.crai': cram_obj['output'] + '.crai'}
         )
-        cpg_sample_id = cram_obj['id']
 
         # Working with CRAM files requires the reference fasta
         ref = b.read_input_group(
--- a/str/runners/str_iterative_eh_runner.py
+++ b/str/runners/str_iterative_eh_runner.py
@@ -3,9 +3,9 @@
 
 """
 This script uses ExpansionHunterv5 to call STRs on WGS cram files.
-Required input: --variant-catalog (file path to variant catalog), --dataset, and external sample IDs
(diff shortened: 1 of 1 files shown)


str/runners/str_iterative_eh_runner.py
```diff
@@ -49,45 +46,56 @@ def main(
...
+        cram_retrieval_query,
+        variables={'dataset': dataset, 'input_cpg_sids': input_cpg_sids},
+    )
+    valid_cpg_ids = []
+    crams_path = []
+    for i in response['project']['sequencingGroups']:
+        valid_cpg_ids.append(i['id'])
+        if (
+            len(i['analyses']) > 1
+        ):  # if there are multiple crams stored per CPG sample ID
```
MattWellie, line 84: should there be an `else` here? It looks like you're not storing anything for SGs with only one CRAM?

hopedisastro: LOL

hopedisastro: yes good pick up

str/runners/str_iterative_eh_runner.py
```diff
@@ -49,45 +46,50 @@ def main(
...
+    for i in response['project']['sequencingGroups']:
+        for cram in i['analyses']:
+            if 'archive' not in cram['output']:  # select the cram that is not archived
+                cram['id'] = i['id']  # ensures cpg id is stored in the dictionary
+                crams_path.append(cram)
+                valid_cpg_ids.append(i['id'])
+                break
+
+    if len(valid_cpg_ids) != len(input_cpg_sids):
+        cpg_sids_without_crams = set(input_cpg_sids) - set(valid_cpg_ids)
```
MattWellie, line 88: Logically this is fine, you could tidy up the logic a little by using a dictionary instead of maintaining a list of sample IDs and a list of CRAM objects:

```suggestion
    crams_by_id = {}
    for i in response['project']['sequencingGroups']:
        for cram in i['analyses']:
            # ignore archived CRAMs
            if 'archive' in cram['output']:
                continue
            crams_by_id['id'] = cram

    if len(crams_by_id) != len(input_cpg_sids):
        cpg_sids_without_crams = set(input_cpg_sids) - set(crams_by_id.keys())
```


This would then require the change a few lines down:

```
for cpg_sample_id, cram_obj in crams_by_id.items():
    ...
```

You won't gain anything functional, but a dictionary is a decent way to
1. keep the SG ID and corresponding CRAM object linked
2. ensure you only have one CRAM per SG ID without inserting breaks in the logic

MattWellie (approved):

PR #84, opened 2023-08-20
Update merge_str_prep.py

hopedisastro:
Main change is switching `external-id` command line argument for an `internal-id`

### DIFF
--- a/str/trtools/merge_str_prep.py
+++ b/str/trtools/merge_str_prep.py
@@ -13,8 +13,6 @@ pip install sample-metadata hail click
 import os
 import click
 
-from sample_metadata.apis import SampleApi
-
 from cpg_utils.config import get_config
 from cpg_utils.hail_batch import output_path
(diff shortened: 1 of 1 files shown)


str/trtools/merge_str_prep.py
```diff
@@ -62,7 +54,7 @@ def main(
 
     input_vcf_dict = {}
 
-    for id in list(external_id_to_cpg_id.values()):
+    for id in list(internal_wgs_ids):
         input_vcf_dict[id] = os.path.join(input_dir, f'{id}_{caller}.vcf')
```
MattWellie, line 58:
```suggestion
    input_vcf_dict = {id: os.path.join(input_dir, f'{id}_{caller}.vcf') for id in internal_wgs_ids}
```

This dict creation can be tidied into a single statement

MattWellie (approved):
Optional syntax change, but fine to approve

PR #85, opened 2023-08-21
Update merge_str_runner.py

hopedisastro:


hopedisastro:
This runs ok on 10 crams. May need more storage when we do the full cohort, but we can figure that out later?

hopedisastro:
combined zip and indexing with the mergeSTR command and confirms it runs ok

str/trtools/merge_str_runner.py
```diff
@@ -33,42 +31,51 @@
...
-    output_path_vcf = output_path(f'mergeSTR_{num_samples}_samples_eh.vcf.bgz')
-    output_path_vcf = output_path_vcf.removeprefix('gs://').split('/', maxsplit=1)[1]
-
     trtools_job.command(
         f"""
-    mergeSTR --vcfs {vcffuse_path} --out temp.vcf --vcftype eh
-    bgzip -c temp.vcf > /vcffuse/{output_path_vcf}
-    tabix /vcffuse/{output_path_vcf}
+    mergeSTR --vcfs {vcffuse_path} --out {trtools_job.vcf_output} --vcftype eh
+    bgzip -c {trtools_job.vcf_output}.vcf > {trtools_job.vcf_output['vcf.gz']}
```
MattWellie, line 68: This works fine, so just a nit here. 

You've made a `vcf` entry in the resource group, so you can use `{trtools_job.vcf_output['vcf'}` here to be consistent. 

Alternatively you could drop `vcf` from the resource group, then below when you have `b.write_output(..)` you can simplify this to

```
b.write_output(trtools_job.vcf_output, output_path_name)
```

which would automatically copy out the `vcf.gz` and `vcf.gz.tbi`

MattWellie (approved):

PR #87, opened 2023-08-24
skip over ONT bams, make sure null entry doesn't cause error

hopedisastro:


str/runners/str_iterative_eh_runner.py
```diff
@@ -85,8 +85,13 @@ def main(
...
         for cram in i['analyses']:
-            # ignore archived CRAMs
-            if 'archive' in cram['output']:
+            # ignore archived CRAMs and ONT crams
+            try:
+                if 'archive' in cram['output']:
+                    continue
+                elif 'ont' in cram['output']:
+                    continue
+            except:
```
MattWellie, line 94: Naked exceptions are not advised as they'll hide any error (including the ones you aren't expecting)

You could replace this try;except with `if not cram['output']: continue`, or `if cram['output'] is None: continue`

MattWellie (approved):

PR #93, opened 2023-11-25
Sharded version of current hipstr script: str_iterative_hipstr_runner_sharded.py

hopedisastro:
Allows for parallelisation by sharding a big catalog into mini catalogs.

### DIFF
--- a/str/runners/str_iterative_hipstr_runner_sharded.py
+++ b/str/runners/str_iterative_hipstr_runner_sharded.py
@@ -1,0 +1,172 @@
+#!/usr/bin/env python3
+
+
+"""
+This script uses HipSTR to call STRs on WGS cram files, using the joint calling option, and a sharded catalog.
+For example:
+analysis-runner --access-level test --dataset tob-wgs --description 'hipstr run' --output-dir 'str/sensitivity-analysis/hipstr' str_iterative_hipstr_runner.py  --output-file-name=hipster_90_genomes  --variant-catalog=gs://.... --dataset=hgdp HGDP00511
+
+Required packages: sample-metadata, hail, click, os
+pip install sample-metadata hail click
+
+"""
+import logging
+
+import click
+
+from metamist.graphql import gql, query
+from cpg_utils.config import get_config
+from cpg_utils.hail_batch import output_path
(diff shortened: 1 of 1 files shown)


hopedisastro:
Planning to experiment with different storage and memory parameters with larger set of crams beyond the test dataset, so added them as command line params.

str/runners/str_iterative_hipstr_runner_sharded.py
```diff
@@ -0,0 +1,172 @@
...
+
+        hipstr_output_path_viz = output_path(
+            f'{output_file_name}_shard{i+1}.viz.gz', 'analysis'
+        )
+        b.write_output(hipstr_job.hipstr_output['viz.gz'], hipstr_output_path_viz)
+
+        hipstr_output_path_log = output_path(
+            f'{output_file_name}_shard{i+1}.log.txt', 'analysis'
+        )
+        b.write_output(hipstr_job.hipstr_output['log.txt'], hipstr_output_path_log)
```
MattWellie, line 166: I think you can rewrite this whole section as
'''
b.write_output(hipstrjob.hipstroutput, f'...')
'''

This will copy the whole output group out to the same root, with all the correct extensions as defined in the resource group

MattWellie (changes requested):
A few easy simplifications, especially around writing out. Get_batch has been migrated to cpgutils and the version you're importing won't exist soon. 

Sorry for formatting issues, on mobile

str/runners/str_iterative_hipstr_runner_sharded.py
```diff
@@ -0,0 +1,163 @@
...
+    ref = b.read_input_group(
+        **dict(
+            base=ref_fasta,
+            fai=ref_fasta + '.fai',
+            dict=ref_fasta.replace('.fasta', '').replace('.fna', '').replace('.fa', '')
+            + '.dict',
+        )
+    )
+
+    catalog_files = list(to_path(variant_catalog).glob('*.bed'))
```
MattWellie, line 117: you might be able to run enumerate on the results of the `glob` directly, but this is fine

MattWellie (approved):
LGTMatt

PR #95, opened 2023-12-13
Create indel_locus_extractor.py

hopedisastro:
This script extracts the coordinates of every indel recorded in a cohort's Hail matrix table. 

Trying out an experiment to see if I can use indels as a proxy for a potential STR by intersecting these output indel coordinates with my STR catalog.

### DIFF
--- a/str/indel_locus_extractor.py
+++ b/str/indel_locus_extractor.py
@@ -18,7 +18,6 @@ from cpg_utils.hail_batch import output_path, init_batch
 from cpg_workflows.batch import get_batch
 
 
-
 def indel_coordinate_extractor(file_path, gcs_path):
     init_batch()
     mt = hl.read_matrix_table(file_path)
--- a/str/indel_locus_extractor.py
+++ b/str/indel_locus_extractor.py
@@ -39,7 +39,6 @@ def indel_coordinate_extractor(file_path, gcs_path):
 )
 @click.command()
 def main(file_path):
-
     gcs_output_path = output_path(f'indels.tsv', 'analysis')
     indel_coordinate_extractor(file_path, gcs_output_path)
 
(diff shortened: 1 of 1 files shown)


hopedisastro:
I have run `black` on my end and it doesn't change the script, so not sure why the lint check is failing at `black`.

str/indel_locus_extractor.py
```diff
@@ -0,0 +1,60 @@
...
+
+    b = get_batch()
+    hail_job = b.new_python_job(name=f'Hail query filter for indels')
+    hail_job.image(config['workflow']['driver_image'])
+    hail_job.storage('20G')
+    hail_job.cpu(4)
+    gcs_output_path = output_path(f'indels.tsv', 'analysis')
+    hail_job.call(indel_coordinate_extractor, file_path, gcs_output_path)
+
+    b.run(wait=False)
```
MattWellie, line 56:
```suggestion
    gcs_output_path = output_path(f'indels.tsv', 'analysis')
    indel_coordinate_extractor(file_path, gcs_output_path)
```

You're already running this inside the driver image in hail, you should be fine running this without a second layer of batching

str/indel_locus_extractor.py
```diff
@@ -0,0 +1,60 @@
...
+
+import hail as hl
+import click
+
+from cpg_utils.config import get_config
+from cpg_utils.hail_batch import output_path, init_batch
+from cpg_workflows.batch import get_batch
+
+
+config = get_config()
```
MattWellie, line 22:
```suggestion
import hail as hl
import click

from cpg_utils.hail_batch import output_path, init_batch
from cpg_workflows.batch import get_batch

```

If you don't create a nested batch job you don't need the config here

hopedisastro:
Ah, I had to update my `black` version!

MattWellie (approved):

PR #98, opened 2024-01-03
Create vcf_combiner.py

hopedisastro:
EH 5M loci run was done using 50 sharded subcatalogs, meaning that each sample has 50 sharded VCFs. 
This script combines the 50 sharded VCFs into one combined VCF (ie each sample will have one combined VCF for downstream use). 

This script assumes that the 
- ##fileformat, ##FORMAT, ##INFO, ##FILTER lines are constant across all sharded VCFs, and therefore extracts these lines from the first shard to write to output. 
- ##ALT lines are collected as a set across all sharded VCFs (set structure removes duplicates), and then sorted alphabetically to be written to output. 
-  remaining lines are calls (genotypes) at each locus, and are added sequentially to the output file. [VCF sort is run as a downstream step of the pipeline so no need to sort the loci by coordinates now].

### DIFF
--- a/str/helper/vcf_combiner.py
+++ b/str/helper/vcf_combiner.py
@@ -2,7 +2,7 @@
 # pylint: disable=import-error, too-many-locals
 
 """
-This script combines sharded VCFs (from one sample and one caller) into one VCF.
+This script combines sharded ExpansionHunter VCFs (from one sample and one caller) into one VCF.
 
 analysis-runner --access-level test --dataset tob-wgs --description \
     'VCF combiner' --output-dir 'str/5M_run_combined_vcfs' \
@@ -19,7 +19,7 @@ config = get_config()
 
 
 def combine_vcf_files(input_dir, gcs_out_path):
-    """Combines sharded VCFs in input_dir into one combined VCF, writing it to a GCS output path"""
+    """Combines sharded ExpansionHunter VCFs in input_dir into one combined VCF, writing it to a GCS output path"""
     input_files = to_path(input_dir).glob('*.vcf')
 
     # Initialize variables to store information
--- a/str/helper/vcf_combiner.py
+++ b/str/helper/vcf_combiner.py
@@ -29,10 +29,10 @@ def combine_vcf_files(input_dir, gcs_out_path):
     chrom_line = ''
     gt_lines = []
 
-    shard_counter =0
+    shard_counter = 0
     # Process each input file
     for file_index, input_file in enumerate(input_files):
-        shard_counter= shard_counter+1
+        shard_counter = shard_counter + 1
         with to_path(input_file).open() as f:
             for line in f:
                 # Collect information from the header lines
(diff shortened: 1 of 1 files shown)


hopedisastro:
Can look at shards in `test-analysis` here: `gs://cpg-tob-wgs-main-analysis/str/5M_run/`

str/helper/vcf_combiner.py
```diff
@@ -0,0 +1,88 @@
...
+        with to_path(input_file).open() as f:
+            for line in f:
+                # Collect information from the header lines
+                if line.startswith('##fileformat') and file_index == 0:
+                    fileformat_line = line
+                elif (
+                    line.startswith('##INFO')
+                    or line.startswith('##FILTER')
+                    or line.startswith('##FORMAT')
+                ) and file_index == 0:
```
MattWellie, line 43: won't you also need the `##contig` lines?

str/helper/vcf_combiner.py
```diff
@@ -0,0 +1,88 @@
...
+                    # Collect ALT lines from all files into a set to remove duplicates
+                    alt_lines.add(line)
+                elif line.startswith('#CHROM') and file_index == 0:
+                    chrom_line = line
+                else:
+                    # Collect calls after #CHROM
+                    gt_lines.append(line)
+
+    # Sort ALT lines alphabetically and convert to a list
+    sorted_alt_lines = sorted(list(alt_lines))
```
MattWellie, line 55:
```suggestion
    sorted_alt_lines = sorted(alt_lines)
```

afaik sorting a set results in a list

str/helper/vcf_combiner.py
```diff
@@ -0,0 +1,91 @@
...
+        # Write CHROM line
+        out_file.write(chrom_line)
+        # Write GT or lines containing the calls
+        out_file.writelines(gt_lines)
+
+
+@click.command()
+@click.option('--input-dir', help='Input directory for sharded VCFs')
+def main(input_dir):
+    # pylint: disable=missing-function-docstring
```
MattWellie, line 84: It's surely easier to add a docstring than ignore the lint error 😓

```suggestion
    """
    Takes an input directory containing vcf shards
    Aggregates all sharded data into a single output file
    """
```

str/helper/vcf_combiner.py
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env python3
+# pylint: disable=import-error, too-many-locals
```
MattWellie, line 2: Do you know what is causing these errors? And do they definitely still occur? It doesn't seem like you have any import errors. The local variables count is quite low by default, but it's been overridden as 25 [here](https://github.com/populationgenomics/sv-workflows/blob/main/.pylintrc#L43), and you have way fewer than that

MattWellie (approved):
Nits about ignoring pylint errors as standard - you should always try and address these issues by default, not just get used to ignoring them. If there are necessary ones to ignore, it might be worth updating [here](https://github.com/populationgenomics/sv-workflows/blob/main/.pylintrc) e.g. `no-value-for-parameter` which is a rubbish side effect of Click.

str/helper/vcf_combiner.py
hopedisastro: sorry, this was copy and pasted from an old script, deleting now

PR #101, opened 2024-01-04
Vcfcombiner sort patch

hopedisastro:


hopedisastro:
The error was that certain header lines were being written out multiple times eg ##fileformat and ##INFO. corrected the VCF parser function.

hopedisastro:
will do the sorting of the calls by locus coordinates in the downstream script involving `bcftools sort`

MattWellie (approved):
LGTM

PR #102, opened 2024-01-05
merge str patch

hopedisastro:
- add bcftools sort 
- update ref fasta

### DIFF
--- a/str/trtools/merge_str_prep.py
+++ b/str/trtools/merge_str_prep.py
@@ -73,7 +73,7 @@ def main(
             bcftools_job.command(
                 f"""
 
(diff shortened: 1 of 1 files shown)


str/trtools/merge_str_prep.py
```diff
@@ -61,28 +59,25 @@ def main(
     for id in list(input_vcf_dict.keys()):
         bcftools_job = b.new_job(name=f'{id} {caller} Files prep')
         bcftools_job.image(BCFTOOLS_IMAGE)
-        bcftools_job.storage('20G')
-        bcftools_job.cpu(8)
+        bcftools_job.cpu(4)
+        bcftools_job.storage('15G')
```
MattWellie, line 63: Is this going to be enough for the full joint-call, or is this stage still per-sample?

MattWellie (approved):

str/trtools/merge_str_prep.py
hopedisastro: still per sample

PR #103, opened 2024-01-05
update import statements, bump up cpu/memory

hopedisastro:
update import statements for get_batch and output_path.

### DIFF
--- a/str/trtools/merge_str_runner.py
+++ b/str/trtools/merge_str_runner.py
@@ -1,10 +1,10 @@
 #!/usr/bin/env python3
 """
-This script merges ExpansionHunter vcf.gz files into one combined VCF. 
+This script merges ExpansionHunter vcf.gz files into one combined VCF.
 Please ensure merge_prep.py has been run on the vcf files prior to running mergeSTR.py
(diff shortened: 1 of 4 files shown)


str/trtools/merge_str_runner.py
```diff
@@ -42,7 +41,7 @@ def main(
     # Initialise TRTools job to run mergeSTR
     trtools_job = b.new_job(name='mergeSTR')
     trtools_job.image(TRTOOLS_IMAGE)
-    trtools_job.cpu(8)
+    trtools_job.cpu(16)
     # mount using cloudfuse for reading input files
     trtools_job.cloudfuse(f'cpg-{dataset}-main-analysis', '/vcffuse')
```
MattWellie, line 46: the more robust way to grab the analysis bucket name would be `config.storage.DATASET.analysis` - if we ever change storage patterns (or if you want to run on GCP & Azure without code changes) the config would be the best place to draw from. 

This is fine, wouldn't bother changing here, just an FYI

MattWellie (approved):
LGTM

PR #106, opened 2024-01-08
Add shard compatibility to mergeSTR 

hopedisastro:
Main changes (similar to changes to merge_str_prep.py): 
- Add a 'numshards' parameter to specify number of shards to loop over. 
- Push mergeSTR into a loop over each shard so that mergeSTR runs on all samples on each shard separately.


hopedisastro:
Remaining lint conflicts relate to a different file in the repo, so have been ignored.

PR #107, opened 2024-01-09
Create bcftools_merge_runner.py

hopedisastro:
`bcftools merge` seems to be an OK (and faster) alternative to `mergeSTR` (which has been stalling with larger catalogs see: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1704685344162619).

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -19,7 +19,8 @@
 # 12. redefined-builtin
 # 13. arguments-renamed
 # 14. import-error
-# 14. R0801: Similar lines in 2 files/duplicate-code
+# 15. R0801: Similar lines in 2 files/duplicate-code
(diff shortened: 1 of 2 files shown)


hopedisastro:
disabling `unsubscriptable-object` lint check because it's throwing errors for other files in the repo that are not included in this PR

hopedisastro:
Not sure how to globally disable `unscriptable-object`

Test run on 2 samples: https://batch.hail.populationgenomics.org.au/batches/431698 used 1Gb of storage; so anticipate storage must be at least 50Gb for 100 samples, have added a storage param so we can modify it if we use this script to merge different number of samples.

MattWellie (approved):

PR #109, opened 2024-01-10
Remove gcsfuse from mergeSTR script 

hopedisastro:
- removes gcfuse from mergeSTR script, replace with local file copying into job environment
- adds job storage param to change when needed.


hopedisastro:
the tortoise (mergeSTR) beats the hare (bcftools merge)...

MattWellie (approved):
🐢 > 🐇

PR #110, opened 2024-01-10
Create `joint_merge_str_runner.py`

hopedisastro:
This mergeSTR script is able to merge STR VCFs across 2 file directories, or dataset buckets. 

Relative to existing `merge_str_runner.py`, this script requires: 
- two input file directories
- two sample ID lists (containing internal sample IDs separated by \n)


hopedisastro:
(will wait until mergeSTR batch job on a single directory successfully completes before requesting review)

PR #113, opened 2024-01-11
Step 3 of debug: Create `merge_str_vcf_combiner.py`

hopedisastro:
This is a new script that concatenates sharded mergeSTR output VCFs together into one combined VCF. 

Please note there is a custom sorted order of shards because shard_label of the VCFs unintentionally followed a different order to the way the catalog was sharded; so I am enforcing this order to ensure the VCF shards are added in the right chromosome/position order. 

Details of the shard order error: 
Catalog shards were read in as shard_1, shard_10, shard_11...shard_19, shard_2, shard_20..and so on (alphabetical order) 
This means that catalog shard_10 maps to VCF shard_2 because when EH was running, the shard_counter just increased by 1 each time a new catalog shard was read in. I will fix this after the BioHEART n=100 run (just so that this custom sort order applies to all n=200 samples).

### DIFF
--- a/str/helper/merge_str_vcf_combiner.py
+++ b/str/helper/merge_str_vcf_combiner.py
@@ -1,0 +1,125 @@
+#!/usr/bin/env python3
+
+"""
+This script combines sharded output VCFs from mergeSTR into one VCF, and assumes the genotyper used was ExpansionHunter.
+
+analysis-runner --access-level test --dataset tob-wgs --description \
+    'VCF combiner' --output-dir 'hoptan-str/shard_workflow_test/merge_str_vcf_combiner' \
+    merge_str_vcf_combiner.py \
+    --input-dir=gs://cpg-tob-wgs-test/hoptan-str/shard_workflow_test/merge_str \
+"""
+import click
+
+from cpg_utils.config import get_config
+from cpg_utils import to_path
+from cpg_utils.hail_batch import get_batch, output_path
+
+import gzip
+
+config = get_config()
+
+
+def combine_vcf_files(input_dir, gcs_out_path):
+    """Combines sharded mergeSTR output VCFs in input_dir into one combined VCF, writing it to a GCS output path"""
+
+    input_file_paths = to_path(input_dir).glob('*.vcf.gz')
+
+    # make input_files GSPath elements into a string type object
+    input_file_paths = {str(gs_path) for gs_path in input_file_paths}
+
+    #create a dictionary where key is the shard number and value is the path to the shard
+    input_files_dict = {}
+    for file_path in input_file_paths:
+        key = int(file_path.split('eh_shard')[1].split('.')[0])
+        input_files_dict[key] = file_path
+
+    #custom sorted order of shards - PLEASE UPDATE FOR LATER USE
+    sorted_key_order  = [
+    1, 12, 23, 34, 45, 47, 48, 49, 50,
+    2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 13,
+    14, 15, 16, 17, 18, 19, 20, 21, 22, 24,
+    25, 26, 27, 28, 29, 30, 31, 32, 33, 35,
+    36, 37, 38, 39, 40, 41, 42, 43, 44, 46
+]
+
+    # Initialize variables to store information
+    fileformat_line = ''
+    info_lines = []
+    alt_lines = set()
+    chrom_line = ''
+    gt_lines = []
+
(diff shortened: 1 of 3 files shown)


hopedisastro:
Confirming this produces the number of records we expect from genotyping the entire catalog!

str/helper/merge_str_vcf_combiner.py
```diff
@@ -0,0 +1,168 @@
...
+    input_file_paths = {str(gs_path) for gs_path in input_file_paths}
+
+    # create a dictionary where key is the shard number and value is the path to the shard
+    input_files_dict = {}
+    for file_path in input_file_paths:
+        key = int(file_path.split('eh_shard')[1].split('.')[0])
+        input_files_dict[key] = file_path
+
+    # custom sorted order of shards - PLEASE UPDATE FOR LATER USE
+    sorted_key_order = [
```
MattWellie, line 36: I am very confused by this ordering...

str/helper/merge_str_vcf_combiner.py
```diff
@@ -0,0 +1,168 @@
...
+    # Initialize variables to store information
+    fileformat_line = ''
+    info_lines = []
+    alt_lines = set()
+    chrom_line = ''
+    gt_lines = []
+
+    shard_counter = 0
+    # Process each input file
+    for key in sorted_key_order:
```
MattWellie, line 98:
```suggestion
    for shard_counter, key in enumerate(sorted_key_order):
```

str/helper/merge_str_vcf_combiner.py
```diff
@@ -0,0 +1,168 @@
...
+        input_file = to_path(input_files_dict[key])
+        print(f'Parsing {input_file}')
+        shard_counter = shard_counter + 1
```
MattWellie, line 101:
```suggestion
```

str/helper/merge_str_vcf_combiner.py
```diff
@@ -0,0 +1,168 @@
...
+                    # Collect ALT lines from all files into a set to remove duplicates
+                    alt_lines.add(line)
+                elif line.startswith('#CHROM'):
+                    if key == 1:
+                        chrom_line = line
+                elif not line.startswith('#'):
+                    # Collect calls after #CHROM
+                    gt_lines.append(line)
+
+    print(f'Parsed {shard_counter} sharded VCFs')
```
MattWellie, line 128:
```suggestion
    print(f'Parsed {len(sorted_key_order)} sharded VCFs')
```

str/helper/merge_str_vcf_combiner.py
```diff
@@ -0,0 +1,168 @@
...
+    """
+
+    # Initializing Batch
+    b = get_batch()
+
+    combiner_job = b.new_python_job(name=f'VCF Combiner job')
+    out_path = output_path(f'combined_eh.vcf', 'analysis')
+    combiner_job.call(combine_vcf_files, input_dir, out_path)
+
+    b.run(wait=False)
```
MattWellie, line 164:
```suggestion
    out_path = output_path(f'combined_eh.vcf', 'analysis')
    combine_vcf_files(input_dir, out_path)
```

No need for this - we're already running a python task in a runtime which has all the auth required to get on with the job. Makes no sense for the whole purpose of this Python job to be generating a separate python job in a complex way

str/helper/merge_str_vcf_combiner.py
```diff
@@ -0,0 +1,168 @@
...
+    'VCF combiner' --output-dir 'hoptan-str/shard_workflow_test/merge_str_vcf_combiner' \
+    merge_str_vcf_combiner.py \
+    --input-dir=gs://cpg-tob-wgs-test/hoptan-str/shard_workflow_test/merge_str
+"""
+import gzip
+import click
+
+from cpg_utils.config import get_config
+from cpg_utils import to_path
+from cpg_utils.hail_batch import get_batch, output_path
```
MattWellie, line 16:
```suggestion
from cpg_utils.hail_batch import output_path
```

I don' think there's any need for a nested batch?

MattWellie (changes requested):
AFAIK there's no benefit in spawning a python job - this is already a python process, so just keep it simple

str/helper/merge_str_vcf_combiner.py
hopedisastro: will show in slack DM

MattWellie (approved):
LGTM!

PR #119, opened 2024-01-15
add lightweight option  to EH - output vcf-only

hopedisastro:
Prior discussion suggests that omitting realigned BAM output reduces EH runtime costs. 

Adding `vcf-only` param to test this out

### DIFF
--- a/str/runners/str_iterative_eh_runner.py
+++ b/str/runners/str_iterative_eh_runner.py
@@ -45,7 +45,7 @@ EH_IMAGE = config['images']['expansionhunter_bw2']
 )
 @click.option('--job-memory', help='Memory of the Hail batch job', default='32G')
 @click.option('--job-ncpu', help='Number of CPUs of the Hail batch job', default=8)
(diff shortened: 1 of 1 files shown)


hopedisastro:
We save ~3.50USD per CRAM using the vcf-only option!

str/runners/str_iterative_eh_runner.py
```diff
@@ -45,6 +45,7 @@
 )
 @click.option('--job-memory', help='Memory of the Hail batch job', default='32G')
 @click.option('--job-ncpu', help='Number of CPUs of the Hail batch job', default=8)
+@click.option('--vcf-only', is_flag=True, help='Omits realigned bam and JSON output')
```
MattWellie, line 48: can we go one step further and have this default to True?

hopedisastro: d'accord

MattWellie (approved):
BOOM

PR #127, opened 2024-01-22
Create stat_str_vcf_combiner.py

hopedisastro:
Combines the output of sharded statSTR into one tab output file. (simple file parser)


MattWellie:
As @illusional mentioned on slack it would be good to swap out the temp file/final output file to a compressed format, to reduce storage needs

MattWellie (approved):
Would be good to include file compression on output, and compression of the individual temp files (that's a job for the previous script, if it's too late to regenerate that data in compressed form don't sweat it)

PR #130, opened 2024-01-27
Sorts the sharded catalog files by shard number numerically, not alphabetically! 

hopedisastro:
Previously the shards were sorted alphabetically eg chunk_1, chunk_10, chunk_2 and so on. This change sorts the chunk files numerically. This means that for example job 42 will align with genotyping sites on chunk_42, which is more intuitive.


MattWellie (approved):
LGTM

PR #131, opened 2024-01-27
Removes sorted key order from mergeSTR combiner

hopedisastro:
The sorted key order is not necessary anymore because now the shard_{number} will align with the chunk_{number} after implementing the EH patch.

### DIFF
--- a/str/helper/merge_str_vcf_combiner.py
+++ b/str/helper/merge_str_vcf_combiner.py
@@ -14,61 +14,6 @@ import click
 from cpg_utils import to_path
 from cpg_utils.hail_batch import output_path
 
-# custom sorted order of shards - PLEASE UPDATE FOR LATER USE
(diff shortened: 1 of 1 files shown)


str/helper/merge_str_vcf_combiner.py
```diff
@@ -100,7 +46,7 @@ def main(input_dir, output):
     temporary_gt_file = 'temporary_gt_file.txt'
     with open(temporary_gt_file, 'w', encoding='utf-8') as handle:
         # Process each input file
-        for key in SORTED_KEY_ORDER:
+        for key in sorted(list(input_files_dict.keys())):
```
MattWellie, line 49:
```suggestion
        for key in sorted(input_files_dict.keys()):
```
Just a nit, but sorting any collection will return a list

MattWellie (approved):

PR #133, opened 2024-01-29
Create sex_inference_stand_alone.py

hopedisastro:
This is a standalone script to test out `annotate_sex()`. In particular, we set `variants_only_x_ploidy=True` and
        `variants_only_y_ploidy=False` as per gnomAD v4. 

Much of the script is adapted from `sample_qc.py` in the LC pipeline (some checkpointing has been removed).


hopedisastro:
Successful run: https://batch.hail.populationgenomics.org.au/batches/432416

MattWellie (approved):
Looks useful to debug

PR #136, opened 2024-01-30
revert back to collect()

hopedisastro:
`collect_as_set()` is a Hail aggregator function, so requires some agg. .... to work. 

https://batch.hail.populationgenomics.org.au/batches/432438/jobs/1
`AttributeError: 'StringExpression' object has no attribute 'collect_as_set'
`


MattWellie (approved):
Yeah I boned that one, looking closer at the docs the correct syntax would be 
`hl.agg.collect_as_set(filtered_mt.info.REPID)`

PR #138, opened 2024-01-30
Create copy_catalog_to_tob.sh

hopedisastro:
This script copies catalog files (no sensitive DNA info) from the bioheart bucket into TOB so that it can be used for the TOB genotyping run.


MattWellie:
Looks fine, but should this end up in cpg-common at some point so that it can be a single source for all projects that need it?

MattWellie (approved):
lgtm, should pair this with a separate update of the references config for ease of use (I'll lend a hand)

PR #139, opened 2024-02-01
Add output-file-param to EH runner script

hopedisastro:
When a job fails for one sharded catalog, I will need to manually define the shard number in the output file name


MattWellie (approved):
Rubber stamp!

PR #141, opened 2024-02-02
Update combiner script 

hopedisastro:
Problem - with 890 samples, storage requirements of current script is not feasible (>80G allocated and still failing). 

Changes: 
- Added caching to input files after parsing through each of them one by one - reduces storage
- Used `upload_from()` to upload the temporary output file to GCP instead of writing directly to GCS - reduces memory
- Bgzip all temporary and final files - reduces storage and memory by 4x. 

(https://centrepopgen.slack.com/archives/C030X7WGFCL/p1706846188154669?thread_ts=1705629950.573249&cid=C030X7WGFCL)

### DIFF
--- a/str/helper/merge_str_vcf_combiner.py
+++ b/str/helper/merge_str_vcf_combiner.py
@@ -39,7 +39,7 @@ def main(input_dir, output):
         for file_path in input_file_paths
     }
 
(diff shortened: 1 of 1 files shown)


illusional (approved):
Looks good, ~~can you pad out the approach taken in the PR description a bit more~~ (I see you already did that, never mind)

PR #142, opened 2024-02-05
BGZIP and MT writer

hopedisastro:
Script 1 bgzips a VCF file, writing to output and a separate script reads the bgzip file into a hail matrix table. Outputs Hail matrix table to a GCP bucket, ready for downstream analysis with Hail query.


hopedisastro:
I couldn't get it to work as one script. But I think it's ok to keep it separate like this - allows for greater flexibility in the future

str/helper/bgzip.py
```diff
@@ -0,0 +1,68 @@
...
+            'vcf.bgz.tbi': '{root}.vcf.bgz.tbi',
+        }
+    )
+
+    bcftools_job.command(
+        f"""
+
+        bcftools view {vcf_input} | bgzip -c >  {bcftools_job.vcf_output['vcf.bgz']}
+
+        tabix -f -p vcf {bcftools_job.vcf_output['vcf.bgz']}
```
MattWellie, line 56: -f shouldn't be required here, this is a freshly created name

MattWellie (approved):
LGTM

str/helper/bgzip.py
hopedisastro, line 56:
```suggestion
```

hopedisastro: remove tabix

str/helper/bgzip.py
```diff
@@ -0,0 +1,68 @@
...
+    bcftools_job = b.new_job(name=f'{input_file_name} BGZIPPING')
+    bcftools_job.image(BCFTOOLS_IMAGE)
+    bcftools_job.memory(job_memory)
+    bcftools_job.storage(job_storage)
+    bcftools_job.cpu(job_cpu)
+
+    bcftools_job.declare_resource_group(
+        vcf_output={
+            'vcf.bgz': '{root}.vcf.bgz',
+            'vcf.bgz.tbi': '{root}.vcf.bgz.tbi',
```
hopedisastro, line 47:
```suggestion
```

PR #143, opened 2024-02-05
Exports a MT to GCP bucket

hopedisastro:


str/helper/mt_extractor.py
```diff
@@ -0,0 +1,34 @@
...
+    '--file-path',
+    help='GCS file path to BGZIP VCF file.',
+    type=str,
+)
+@click.command()
+def main(file_path):
+    """writes a BGZIP VCF as a Hail Matrix Table to a GCS bucket"""
+
+    init_batch()
+    gcs_path = output_path('str.mt', 'analysis')
```
MattWellie, line 29: I'm not super wild about this, does all this data need to be in an analysis bucket? I'm mostly concerned that this script becomes a way of taking any VCF data and transcribing it to a less protected bucket

str/helper/mt_extractor.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python3
+# pylint: disable=missing-function-docstring,no-member
```
MattWellie, line 2: I'm not sure either of these are relevant, you have one function with a docstring

str/helper/mt_extractor.py
```diff
@@ -0,0 +1,34 @@
...
+import hail as hl
+import click
+
+from cpg_utils.hail_batch import output_path, init_batch
+
+
+@click.option(
+    '--file-path',
+    help='GCS file path to BGZIP VCF file.',
+    type=str,
```
MattWellie, line 22: Not important, but the default type of a click arg is str, so this shouldn't need clarification

MattWellie (changes requested):

str/helper/mt_extractor.py
hopedisastro, line 2:
```suggestion
```

str/helper/mt_extractor.py
```diff
@@ -0,0 +1,33 @@
...
```
hopedisastro, line 21:
```suggestion
```

str/helper/mt_extractor.py
```diff
@@ -0,0 +1,32 @@
...
+@click.option(
+    '--file-path',
+    help='GCS file path to BGZIP VCF file.',
+)
+@click.command()
+def main(file_path):
+    """writes a BGZIP VCF as a Hail Matrix Table to a GCS bucket"""
+
+    init_batch()
+    gcs_path = output_path('str.mt', 'analysis')
```
hopedisastro, line 27:
```suggestion
    gcs_path = output_path('str.mt')
```

str/helper/mt_extractor.py
hopedisastro: removed analysis from output bucket

str/helper/mt_extractor.py
hopedisastro: removed

str/helper/mt_extractor.py
hopedisastro: removed

MattWellie (approved):
LGTM!

PR #144, opened 2024-02-05
Create mt_joiner.py

hopedisastro:
Script for our niche case where 200 samples had 6M loci genotyped, but the rest of the cohort had only a subset of the loci genotyped. Aim is to throw away the 'extra' loci that were only genotyped in the 200 samples, when creating the combined MT

### DIFF
--- a/str/helper/mt_joiner.py
+++ b/str/helper/mt_joiner.py
@@ -36,7 +36,7 @@ def main(mt_path_1, mt_path_2):
 
     #rekey by REPID
     mt_1 = mt_1.annotate_rows(REPID = mt_1.info.REPID)
-    mt_1 = mt_1.key_rows_by(REPID = mt_1['REPID'], alleles = mt_1['alleles'])
+    mt_1 = mt_1.key_rows_by(REPID = mt_1['REPID'])
 
     print(f'{mt_path_1} dimensions: {mt_1.count()}')
 
@@ -44,7 +44,7 @@ def main(mt_path_1, mt_path_2):
 
     #rekey by REPID
     mt_2 = mt_2.annotate_rows(REPID = mt_2.info.REPID)
-    mt_2 = mt_2.key_rows_by(REPID = mt_2['REPID'], alleles = mt_2['alleles'])
+    mt_2 = mt_2.key_rows_by(REPID = mt_2['REPID'])
 
     print(f'{mt_path_2} dimensions: {mt_2.count()}')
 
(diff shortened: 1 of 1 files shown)


str/helper/mt_joiner.py
```diff
@@ -0,0 +1,59 @@
...
+@click.command()
+def main(mt_path_1, mt_path_2):
+    """joins two Hail matrix tables together and writes combined mt to GCS"""
+
+    init_batch()
+    mt_1 = hl.read_matrix_table(mt_path_1)
+
+    # rekey by REPID
+    mt_1 = mt_1.annotate_rows(REPID=mt_1.info.REPID)
+    mt_1 = mt_1.key_rows_by(REPID=mt_1['REPID'])
```
MattWellie, line 39: This doesn't feel like the right syntax (but maybe it is?). AFAIK key_rows_by takes Strings, so this would be 
```suggestion
    mt_1 = mt_1.annotate_rows(REPID=mt_1.info.REPID)
    mt_1 = mt_1.key_rows_by('REPID')
```

As a side note, do you want to re-key prior to exporting, or do you want this data keyed on REPID?

hopedisastro: I want to keep it keyed by REPID.

hopedisastro: Either syntax is fine :)

str/helper/mt_joiner.py
```diff
@@ -0,0 +1,59 @@
...
+    mt_1 = mt_1.annotate_rows(REPID=mt_1.info.REPID)
+    mt_1 = mt_1.key_rows_by('REPID')
+
+    print(f'{mt_path_1} dimensions: {mt_1.count()}')
+
+    mt_2 = hl.read_matrix_table(mt_path_2)
+
+    # rekey by REPID
+    mt_2 = mt_2.annotate_rows(REPID=mt_2.info.REPID)
+    mt_2 = mt_2.key_rows_by(REPID=mt_2['REPID'])
```
hopedisastro, line 47:
```suggestion
    mt_2 = mt_2.key_rows_by('REPID')
```

hopedisastro:
as a note, because this isn't a proper VCF merge, some of the row fields eg "alleles" will not be reflective of the whole cohort (it will be reflective of the first mt passed in). But I think this is OK because I will not be using these particular fields, and focusing mostly on the entry level fields.

MattWellie (approved):

PR #146, opened 2024-02-12
Create a VCF pruner - filter a VCF for VARIDs that intersect with a provided catalog. 

hopedisastro:
I want to create a cleaner solution for merging the VCFs of the subset of samples with extensive genotyping with the rest of the cohort that had fewer loci genotyped (current solution is doing an ad hoc join operation of two matrix tables, but I don't have a solution yet of creating a nice final VCF containing all the samples from both cohorts). The first step is this PR where we prune the extensively genotyped VCFs to contain the variants that match the provided sharded VCF. This is needed because mergeSTR requires VCF inputs to have the same exact catalog of variants (STRs) genotyped. 

The output of this script is that the VCF is sharded to match the way that the provided sharded VCF is sharded; as sharded VCFs are the expected input for downstream mergeSTR step. 

I have tested the output of this script in a simple example with the downstream mergeSTR - works as intended with no errors. 

I don't know why `black` keeps failing :(

### DIFF
--- a/str/helper/vcf_pruner.py
+++ b/str/helper/vcf_pruner.py
@@ -107,3 +107,6 @@ def main(json_file_dir, vcf_file_dir, cpg_ids: list[str]):
                 out_file.write(chrom_line)
                 # Write GT or lines containing the calls
                 out_file.writelines(gt_lines)
+
+if __name__ == '__main__':
+    main()  # pylint: disable=no-value-for-parameter
--- a/str/helper/vcf_pruner.py
+++ b/str/helper/vcf_pruner.py
@@ -26,11 +26,28 @@ def variant_id_collector(catalog_file):
     """Collects all variant IDs from a sharded catalog file and returns a set of unique variant IDs."""
 
     variant_ids = []
-    fileformat_line = ''
-    info_lines = []
 
     with to_path(catalog_file).open() as f:
+        for line in f:
+            if not line.startswith('#'):
+                row_info = line.split('\t')[7]
+                var_id = (row_info.split(';')[4])[6:]
+                variant_ids.append(var_id)
 
+    return set(variant_ids), variant_ids
+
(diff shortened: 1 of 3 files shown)


hopedisastro:
Test batch run: https://batch.hail.populationgenomics.org.au/batches/433094

str/helper/vcf_pruner.py
```diff
@@ -0,0 +1,155 @@
...
+                row_info = line.split('\t')[7]
+                var_id = {(row_info.split(';')[4])[6:]}
+
+                if (
+                    var_id & variant_id_set
+                ):  # if the variant ID is in the set of target variant IDs
+                    var_id = ''.join(
+                        map(str, var_id)
+                    )  # converts var_id from set to string
+                    gt_lines[var_id] = line
```
MattWellie, line 80: Does each row have a single variant ID, or multiple? The check of 'is a string in a set of strings' is faster than 'does this set of strings intersect with this set'. I pulled an example file and the VARID looks singular
```END=1234;REF=1;RL=12;RU=A;VARID=1-1122-1234-A;REPID=1-1122-1234-A```

```suggestion
                row_info = line.split('\t')[7]
                var_id = row_info.split(';')[4].removeprefix('VARID=')

                # if the variant ID is in the set of target variant IDs
                if var_id in variant_id_set:
                    gt_lines[var_id] = line
```

MattWellie: this could be more terse, but would need checking 
```
if var_id := row_info.split(';')[4].removeprefix('VARID=') in variant_id_set:
    gt_lines[var_id] = line
```

MattWellie (approved):
the set intersection doesn't seem necessary to me, but will work. The lack of inter-job dependencies means that this could start a whole bunch of jobs at once, so be wary of that. I've no idea what number of shards/samples this will run for

PR #147, opened 2024-02-13
mergeSTR update - add option to include files from up to four different file directories

hopedisastro:
Update mergeSTR runner so that it can: 
- accept files from up to four different file directories (previously only max of two file directories were possible): am expecting two distinct file directories for each of the two cohorts = 4 file directory sources
-  receive memory, cpu params for the Hail batch job from the user
I disabled =`too-many-arguments`,`too-many-locals`

### DIFF
--- a/str/trtools/merge_str_runner.py
+++ b/str/trtools/merge_str_runner.py
@@ -6,7 +6,8 @@ Please ensure merge_prep.py has been run on the vcf files prior to running merge
 Optional ability to add in VCFs from another file directory (but must be sharded in the same way as the input-dir-1)
 
 For example:
-analysis-runner --access-level standard --dataset tob-wgs --description '5M merge TOB100' --output-dir 'str/5M_run_combined_vcfs/merge_str/v4' merge_str_runner.py --input-dir-1=gs://cpg-tob-wgs-main-analysis/str/5M_run_combined_vcfs/merge_str_prep/v4-2 --num-shards=50 --sample-list-1=gs://
+analysis-runner --access-level full --dataset tob-wgs --description '5M-3M mergeSTR tester' --output-dir 'str/5M-3M experiment/merge_str/v1' merge_str_runner.py --input-dir-1=gs://cpg-tob-wgs-main-analysis/str/5M_run_combined_vcfs_pruned/merge_str_prep/v4 --num-shards=27 \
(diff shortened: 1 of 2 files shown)


str/trtools/merge_str_runner.py
```diff
@@ -102,6 +128,39 @@ def main(
                         )['vcf.gz']
                     )
                 num_samples = num_samples + len(ids_2)
+        # similarly, if third and fourth file directories are specified - read in input file paths:
```
MattWellie, line 131: Can we rewrite the inputs/logic here? Copy pasting the code arbitrarily could be replaced with something more elegant. I was thinking you could replace this `input1 sample1 input2 sample2...` logic with an argument that takes an arbitrary number of strings as arguments `sample1,input1 sample2,input2`, then iterate over those as 
```
for sample, input in arg.split(','):
    do the stuff
```

which is not elegant, but does mean that the same code will handle any number of inputs without rewriting the logic

MattWellie: Also are the sample IDs expected to be different for each directory? I'm not super familiar with this process, but I assumed all the co-processed files would be shards within a consistent callset :/

hopedisastro: Thanks, yes sample IDs _**should**_ be different for each directory. I should probably add a warning if duplicate IDs are found

str/trtools/merge_str_runner.py
```diff
@@ -95,46 +72,14 @@ def main(
...
-                        )['vcf.gz']
-                    )
-                num_samples = num_samples + len(ids_2)
-        # similarly, if third and fourth file directories are specified - read in input file paths:
-        if input_dir_3 is not None:
-            with to_path(sample_list_3).open() as f_3:
-                ids_3 = [line.strip() for line in f_3]
-                for id in ids_3:
+        cpg_ids =[]
+        for input_dir, sample_list in zip(input_file_paths[::2], input_file_paths[1::2]):
```
MattWellie: I'm not sure this logic works... in your example command up top you have one continuous string with commas, as
`dir1,sample1,dir2,sample2`. You would need to break that into a list, then iterate over the list in pairs. This logic would do a staggered iteration over a continuous list, but first you need to break it into a list of separate elements.

With `nargs=-1` in your click argument description you can ingest any number of strings into a list, which you can accommodate by making the args `dir1,sample1    dir2,sample2`. If you comma-separate the directory and sample list, then separate each pair with spaces, you end up with

```
script.py ... dir1,sample1    dir2,sample2
...
input_file_paths = ['dir1,sample1', 'dir2,sample2']
for dir, sam in input_file_paths.split(','): 
   ...
```

Alternatively your logic as currently written will work, but you need to remove your commas from the example invocation up top so that each of the strings is kept separately in a continuous list

hopedisastro: i changed it to follow the args convention of dir1,sample1    dir2,sample2 - pairs are separated by space.

MattWellie (approved):

PR #149, opened 2024-02-13
Increase QoB memory using 'highmem' option

hopedisastro:
Increase QoB memory using 'highmem' option

Reference: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1707187236506529
Failed QoB job: https://batch.hail.populationgenomics.org.au/batches/433177


annacuomo (approved):
LGTM

PR #150, opened 2024-02-19
Update file paths to be relative to $BATCH_TMP/inputs

hopedisastro:
Try to overcome the `Argument too long` error from mergeSTR because the VCF list passed in at the command line is getting quite long https://centrepopgen.slack.com/archives/C030X7WGFCL/p1708146312785919

### DIFF
--- a/str/trtools/merge_str_runner.py
+++ b/str/trtools/merge_str_runner.py
@@ -95,8 +95,12 @@ def main(
         if len(cpg_ids) != len(set(cpg_ids)):
             raise ValueError('Duplicate CPG IDs detected in sample list')
 
+        #make batch_vcfs relative to BATCH_TMP/inputs
+        batch_vcfs = [vcf.split('inputs/')[1] for vcf in batch_vcfs]
+
         trtools_job.command(
             f"""
+        cd $BATCH_TMP/inputs
         mergeSTR --vcfs {",".join(batch_vcfs)} --out {trtools_job.vcf_output} --vcftype eh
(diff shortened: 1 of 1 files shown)


MattWellie:
I'm not sure if this will work, but can't help to try...? 

The error text is `Argument list too long`, and the list here is the same length but the contents are slightly smaller. It's possible that this will solve the issue, but if the list length is the real issue that remains unchanged

jmarshall:
I misspoke on Slack; it's actually `$BATCH_TMPDIR`.

However this approach won't work because Hail uses placeholders in variables like `batch_vcfs` and does not instantiate or expand them until very late in the piece (when the eventual command is written out), so `batch_vcfs` entries do not actually contain `inputs/` yet at the point it would be possible to manipulate them in Python. So this has to be done in a more messy way instead — see #151.

str/trtools/merge_str_runner.py
```diff
@@ -95,8 +95,12 @@ def main(
         if len(cpg_ids) != len(set(cpg_ids)):
             raise ValueError('Duplicate CPG IDs detected in sample list')
 
+        # make batch_vcfs relative to BATCH_TMP/inputs
+        batch_vcfs = [vcf.split('inputs/')[1] for vcf in batch_vcfs]
```
MattWellie, line 99: I'm not sure if you can run a `.split()` on a ResourceFile directly, or whether you'll need to do that on the String representation of the same... It may work?
```suggestion
        batch_vcfs = [str(vcf).split('inputs/')[1] for vcf in batch_vcfs]
```

PR #151, opened 2024-02-19
Use relative paths for mergeSTR input file arguments

jmarshall:
When merging thousands of VCFs, using absolute paths here exceeded the shell's command line length limit. As Hail's input filenames match `${BATCH_TMPDIR}/inputs/…` we can use relative paths by trimming off the leading `${BATCH_TMPDIR}` and so avoid the length limit for longer.

Because Hail uses placeholders in variables like `batch_vcfs` and does not instantiate or expand them until very late in the piece (when the eventual command is written out), we need to do this explicitly in the shell snippet rather than earlier (and more clearly 😢) in Python.

In future, mergeSTR may have a `--vcfs-list` file-list option. When that is released, we will be able to recode this to build a temporary file listing the VCF file names instead.

Context: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1708146312785919

### DIFF
--- a/str/trtools/merge_str_runner.py
+++ b/str/trtools/merge_str_runner.py
@@ -95,9 +95,11 @@ def main(
         if len(cpg_ids) != len(set(cpg_ids)):
             raise ValueError('Duplicate CPG IDs detected in sample list')
 
+        # TODO Convert to use mergeSTR --vcfs-list when that option is available
         trtools_job.command(
             f"""
-        mergeSTR --vcfs {",".join(batch_vcfs)} --out {trtools_job.vcf_output} --vcftype eh
+        cd ${{BATCH_TMPDIR}}
+        mergeSTR --vcfs `(echo {";echo ,".join(batch_vcfs)}) | sed 's|'${{BATCH_TMPDIR}}'/*||' | tr -d '\\n'` --out {trtools_job.vcf_output} --vcftype eh
         bgzip -c {trtools_job.vcf_output}.vcf > {trtools_job.vcf_output['vcf.gz']}
         tabix -f -p vcf {trtools_job.vcf_output['vcf.gz']}  > {trtools_job.vcf_output['vcf.gz.tbi']}
         """
--- a/str/trtools/merge_str_runner.py
+++ b/str/trtools/merge_str_runner.py
@@ -98,8 +98,8 @@ def main(
         # TODO Convert to use mergeSTR --vcfs-list when that option is available
         trtools_job.command(
             f"""
-        cd ${{BATCH_TMPDIR}}
-        mergeSTR --vcfs `(echo {";echo ,".join(batch_vcfs)}) | sed 's|'${{BATCH_TMPDIR}}'/*||' | tr -d '\\n'` --out {trtools_job.vcf_output} --vcftype eh
+        cd ${{BATCH_TMPDIR}}/inputs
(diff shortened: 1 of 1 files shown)


MattWellie:
I feel exceptionally unhappy about this as a solution - it feels like we're doing anything to avoid using the appropriate fix, which is picking up the vcfs-list argument. 

The last point at which the MergeSTR code was changed was [here](https://github.com/gymrek-lab/TRTools/tree/f07a99dc51eb3811cd702137fe0c8635b7188ba2), and all tests were passing, even if it wasn't pushed as a formal release. I would target that commit in a docker image, and use the new parameters which solve this issue.

jmarshall:
Me, I would feel exceptionally unhappy about using code that the authors have not released formally without vetting it carefully first. To the point where `--vcfs-list` was merged to their mainline, there are 54 commits affecting TRTools Python code since the 5.0.2 release we are currently using, or 10 commits since the most recent 5.1.1 release. Do we know how those changes may have affected mergeSTR behaviour or the behaviour of other TRTools components that we are using?

To me, this is a sensible workaround until there is an upstream release that contains the `--vcfs-list` option. An almost-as-sensible alternative would be to cherry-pick the `--vcfs-list` commit into a 5.0.2-cpg1 local version that includes the desired option but not other unreleased-by-upstream and unvetted-by-us changes.

str/trtools/merge_str_runner.py
```diff
@@ -95,9 +95,11 @@ def main(
         if len(cpg_ids) != len(set(cpg_ids)):
             raise ValueError('Duplicate CPG IDs detected in sample list')
 
+        # TODO Convert to use mergeSTR --vcfs-list when that option is available
         trtools_job.command(
             f"""
-        mergeSTR --vcfs {",".join(batch_vcfs)} --out {trtools_job.vcf_output} --vcftype eh
+        cd ${{BATCH_TMPDIR}}/inputs
+        mergeSTR --vcfs `(echo {";echo ,".join(batch_vcfs)}) | sed 's|'${{BATCH_TMPDIR}}'/inputs/*||' | tr -d '\\n'` --out {trtools_job.vcf_output} --vcftype eh
```
illusional, line 102: Can this be broken into multiple lines, and potentially set to a variable before hand - this improves the readability of the command line:

```suggestion
        VCFS=$(`$(echo {";echo ,".join(batch_vcfs)}) | sed 's|'${{BATCH_TMPDIR}}'/inputs/*||' | tr -d '\\n'`)
        mergeSTR \
            --vcfs $VCFS \
            --out {trtools_job.vcf_output} \
            --vcftype eh
```

jmarshall: We could (modulo not adding erroneously nested `$(` and backticks, and modulo potential concerns about the maximum size of `VCFS`), but the truth is this is a temporary hack that will go away when we have a TRTools image that includes the `--vcfs-list` option — whichever way that happens.

So I made the minimal change that will evaporate again soon when we change it to `mergeSTR --vcfs-list {batch_vcfs_listfile}` instead, which will be more readable than **main**'s existing `{",".join(batch_vcfs)}` interpolation :shrug:

IMHO the thing that makes this unreadable is the intricacy of the two instances of `echo` due to the crazy comma-formatting needed…

hopedisastro (approved):

PR #153, opened 2024-02-21
Cp Hail Matrix Table containing STR genotypes into `test` bucket 

hopedisastro:
Copies Hail Matrix Table containing STR genotypes into `test` bucket for quick iterative QC plot generation. 

Final QC scripts will be merged into `main`, with final QC plots also produced in respective `main-analysis` buckets. At this point, I will delete the mt from `test`


illusional (approved):
Awesome, thanks!

PR #154, opened 2024-02-23
Add shard param to REViewer runner

hopedisastro:
Adds shard parameter to add in the shard number into the relevant input file paths.


str/runners/reviewer_runner.py
```diff
@@ -26,9 +26,10 @@
 @click.option('--locus', help='Locus identifier as per catalog')
 @click.option('--catalog', help='GCP path to catalog')
 @click.option('--input-dir', help='GCP path to input-dir, includes gs://')
+@click.option('--shard-vcf', help='shard number of the VCF')
```
hopedisastro, line 29:
```suggestion
@click.option('--shard-vcf', help='shard number of the VCF/BAM')
```

MattWellie (approved):
looks like you just need to run Black on this and you're good

PR #157, opened 2024-02-28
Replace `normalize_total` method

hopedisastro:
Changing design of the script so that instead of requiring an Anndata object with the whole genome for all cells (OOM error: https://batch.hail.populationgenomics.org.au/batches/434085/jobs/1), it parses smaller Anndata objects that are subsetted to one chromosome, one cell type. 

This change requires me to DIY the `normalize_total()` method because scanpy's implementation will re-calculate the total counts (representing the number of transcript counts across the whole genome for that particular cell) across the Anndata object. This is incorrect because the Anndata object will not have all the genes anymore (input objects are expected to be restricted to one chromosome). FORTUNATELY, `calculate_qc_metrics()` was run on the giant AnnData object (with all cells, all chromosomes) in an upstream step - storing the true `total_counts` as a row attribute in `obs` (https://github.com/powellgenomicslab/tenk10k_phase1/blob/main/Scanpy/add_metadata_per_sample_no_norm.py#L131), so we can use this attribute during DIY normalization. 

Successful batch: https://batch.hail.populationgenomics.org.au/batches/434123 - please ignore the non-descriptive job name (updated in the most recent commit!)

Still OOM with really large chr1 AnnData objects (rest were successful): https://batch.hail.populationgenomics.org.au/batches/434253/jobs/45 - how do we reduce RAM usage further? 
Tried `backed` mode for the `read_h5ad()` but threw a processing error - I think it has limitations on what processing can be done on backed mode.

### DIFF
--- a/str/associatr/get_cis_numpy_files.py
+++ b/str/associatr/get_cis_numpy_files.py
@@ -1,0 +1,200 @@
+#!/usr/bin/env python3
+# pylint: disable=too-many-arguments,too-many-locals
+
+"""
+This script aims to:
+ - output gene lists for each cell type and chromosome (after filtering out lowly expressed genes)
+ - output cis window files for each gene with scRNA data (cell type + chr specific)
+ - optionally removes samples based on a provided sample file
+ - perform rank-based inverse normal transformation on pseudobulk data (per gene basis)
+ - output gene-level phenotype and covariate numpy objects for input into associatr
+
+ analysis-runner  --config get_cis_numpy_files.toml --dataset "bioheart" --access-level "test" \
+--description "get cis and numpy" --output-dir "str/associatr/rna_pc_calibration/2_pcs/input_files"  \
+--image australia-southeast1-docker.pkg.dev/cpg-common/images/scanpy:1.9.3 \
+python3 get_cis_numpy_files.py
+
+"""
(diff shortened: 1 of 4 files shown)


MattWellie:
It may be a little irritating, but it would definitely be more scalable if you could do something like:

1. read annData object in backed mode
2. write a subset of that to a new object, containing only the fields you need for downstream processing
3. delete the original object and clear up
4. read the smaller version in full mode

Does this sounds feasible at all? I'm not familiar with any part of this process, or the contents of the objects, but they do seem to be unreasonably resource-heavy to use

MattWellie:
Alternatively (?) if you could drop Pandas for [Polars](https://pola.rs/) that could maybe help your scaling?

hopedisastro:
I don't think we can subset the anndata, because we're using scanpy in house tools in particular `regress_out` which needs the full adata object to work. 

What we can do is apply the lowly expressed genes filter upfront, reducing the anndata object size for downstream processes. This job used to fail before, now it's working with ample memory margin: https://batch.hail.populationgenomics.org.au/batches/435257/jobs/1

So I think the OOM concerns are now resolved.

MattWellie (approved):

PR #158, opened 2024-02-28
Create get_covariates.py for associaTR pseudobulk pipeline

hopedisastro:
This script calculates cell-type specific PCs from the pseudobulk RNA data (genome-wide),
merges them with other pre-calculated covariates (from Anna's pipeline), and writes the file to GCP.

`chromosomes` is added as an arg because at this stage not sure if we want to add in chrX or not. So keeping it as an arg for the flexibility. 

Successsful batch - https://batch.hail.populationgenomics.org.au/batches/434148/jobs/1

### DIFF
--- a/str/associatr/get_covariates.py
+++ b/str/associatr/get_covariates.py
@@ -57,7 +57,7 @@ def get_covariates(
 
     # Variance ratio plot
     sc.pl.pca_variance_ratio(adata_genome, save='local.png')
-    hl.hadoop_copy('figures/pca_variance_ratiolocal.png', output_path(f'{cell_type}_scree_plot.png', 'analysis'))
+    hl.hadoop_copy('figures/pca_variance_ratiolocal.png', output_path(f'pseudobulk_RNA_PCs_scree_plots/{cell_type}_scree_plot.png', 'analysis'))
 
     # extract PCs
     df_pcs = pd.DataFrame(adata_genome.obsm['X_pca'])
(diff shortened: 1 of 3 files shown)


str/associatr/get_covariates.py
```diff
@@ -0,0 +1,117 @@
...
+    for i in chromosomes.split(','):
+        gcs_file_path = (
+            f'{pseudobulk_input_dir}/{cell_type}/{cell_type}_chr{i}_pseudobulk.csv'
+        )
+        print(f'Loading {gcs_file_path}...')
+
+        local_file_path = to_path(gcs_file_path).copy('here.csv')
+        adata = sc.read_csv(local_file_path)
+        adatas.append(adata)
+    adata_genome = sc.concat(adatas, axis=1)
```
MattWellie, line 38: If you encounter memory issues this would probably be the spot to include an object deletion
```suggestion
    adata_genome = sc.concat(adatas, axis=1)
    for adata in adatas:
        del adata
    gc.collect()
```

str/associatr/get_covariates.py
```diff
@@ -0,0 +1,117 @@
...
+
+    # read in covariates
+    cov = pd.read_csv(covariate_file_path)
+
+    merged_df = pd.merge(cov, df_pcs, on='sample_id')
+
+    # write to GCP
+    merged_df.to_csv(
+        output_path(f'covariates/{cell_type}_covariates.csv', 'analysis'), index=False
+    )
```
MattWellie, line 69: likewise, this would be a good spot to delete that final AnnData object
```suggestion
    )
    del adata_genome
    gc.collect()
```

MattWellie (approved):
LGTM

PR #159, opened 2024-03-01
Get cis-window and numpy object files - inputs for associaTR

hopedisastro:
This script aims to:
 - output cis window files for each gene with scRNA data (cell type + chr specific)
 - output gene-level phenotype and covariate numpy objects for input into associatr

Successful batch - https://batch.hail.populationgenomics.org.au/batches/434316

### DIFF
--- a/str/associatr/get_cis_numpy_files.py
+++ b/str/associatr/get_cis_numpy_files.py
@@ -12,14 +12,25 @@ import numpy as np
 import pandas as pd
 import scanpy as sc
 
-from cpg_utils.hail_batch import get_batch,output_path
+from cpg_utils.hail_batch import get_batch, output_path
 from cpg_utils import to_path
 
-def cis_window_numpy_extractor(input_h5ad_dir, input_pseudobulk_dir,input_cov_dir,chromosome, cell_type,cis_window, version):
+
+def cis_window_numpy_extractor(
+    input_h5ad_dir,
+    input_pseudobulk_dir,
+    input_cov_dir,
+    chromosome,
+    cell_type,
+    cis_window,
+    version,
+):
     h5ad_file_path = f'{input_h5ad_dir}/{cell_type}_{chromosome}.h5ad'
     expression_h5ad_path = to_path(h5ad_file_path).copy('here.h5ad')
     adata = sc.read_h5ad(expression_h5ad_path)
-    pseudobulk_path = f'{input_pseudobulk_dir}/{cell_type}/{cell_type}_{chromosome}_pseudobulk.csv'
+    pseudobulk_path = (
+        f'{input_pseudobulk_dir}/{cell_type}/{cell_type}_{chromosome}_pseudobulk.csv'
+    )
     pseudobulk = pd.read_csv(pseudobulk_path)
     covariate_path = f'{input_cov_dir}/{cell_type}_covariates.csv'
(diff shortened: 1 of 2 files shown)


str/associatr/get_cis_numpy_files.py
```diff
@@ -0,0 +1,248 @@
...
+    '--min-pct',
+    default=1,
+    help='filter out genes that are expressed in fewer than XX% of cells',
+)
+@click.option(
+    '--remove-samples-file',
+    default=None,
+    help='GCS path to the file containing the list of samples to remove',
+)
+@click.command()
```
MattWellie, line 184: it might be useful to shift to a config-oriented way of passing some arguments, like we did [here](https://github.com/populationgenomics/saige-tenk10k/pull/78) - instead of CLI arguments that have to be passed all over, you can just pop them in a config file, then pull them out at the point they're needed

str/associatr/get_cis_numpy_files.py
```diff
@@ -0,0 +1,248 @@
...
+    cell_types,
+    cis_window,
+    version,
+    job_cpu,
+    job_memory,
+    job_storage,
+    max_parallel_jobs,
+    min_pct,
+    remove_samples_file,
+):
```
MattWellie, line 199: but if you do have arguments, could you please add type hinting
```suggestion
def main(
    input_h5ad_dir: str,
    input_pseudobulk_dir: str,
    input_cov_dir: str,
    chromosomes: str,  # but taken from config should be a list[str]
    cell_types: str,  # but taken from config should be a list[str]
    cis_window:  int,
    version: str,
    job_cpu: int,
    job_memory: str,
    job_storage: str,
    max_parallel_jobs: int,
    min_pct: int,
    remove_samples_file:  str | None,
):
```

str/associatr/get_cis_numpy_files.py
```diff
@@ -0,0 +1,248 @@
...
+import click
+import numpy as np
+import pandas as pd
+import scanpy as sc
+import hail as hl
+import hailtop.batch as hb
+from scipy.stats import norm
+
+
+from cpg_utils.hail_batch import get_batch, output_path, init_batch
```
MattWellie, line 33:
```suggestion
from cpg_utils.hail_batch import get_batch, output_path, init_batch, image_path
```

str/associatr/get_cis_numpy_files.py
```diff
@@ -0,0 +1,248 @@
...
+    for cell_type in cell_types.split(','):
+        for chrom in chromosomes.split(','):
+            init_batch()
+            chrom_len = hl.get_reference('GRCh38').lengths[chrom]
+            j = b.new_python_job(
+                name=f'Extract cis window & phenotype and covariate numpy object for {cell_type}: {chrom}'
+            )
+            j.image(
+                'australia-southeast1-docker.pkg.dev/cpg-common/images/scanpy:1.9.3'
+            )
```
MattWellie, line 225:
```suggestion
            j.image(image_path('australia-southeast1-docker.pkg.dev/cpg-common/images/scanpy:1.9.3'))
```

changing this to pull from config will remove hard coding if you later bump the version in the image

MattWellie (approved):
a few suggestions, happy to workshop the config usage, but nothing major

hopedisastro:
This is now config'ed :)

PR #160, opened 2024-03-03
Create QC_annotator.py

hopedisastro:
This script annotates the ExpansionHunter MT with the following annotations:
- rep_length_alleles: repeat length of each allele
- motif_length: length of the motif
- bp_length_alleles: bp length of each allele
- allele_1_rep_length: repeat length of allele 1
- allele_2_rep_length: repeat length of allele 2
- allele_1_bp_length: bp length of allele 1
- allele_2_bp_length: bp length of allele 2
- aggregated_info.mode_allele:mode allele at each locus
- num_alleles: number of alleles at each locus
- allele_1_minus_mode: difference between allele 1 and mode allele
- allele_2_minus_mode: difference between allele 2 and mode allele
- sum_alleles_is_not_mode: count of non-mode alleles at each locus
- prop_alleles_is_not_mode: proportion of alleles that are not the mode allele per locus
- binom_hwep: binomial Hardy-Weinberg equilibrium p-value
- obs_het: proportion of observed heterozygous calls per locus

Annotations will be used for downstream filtering. 

This runs without issues on the full `mt` already.

### DIFF
--- a/str/qc/qc_annotator.py
+++ b/str/qc/qc_annotator.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 # pylint: disable=missing-function-docstring,no-member
 """
-This script annotates the MT with the following annotations:
+This script annotates the ExpansionHunter MT with the following annotations:
 - rep_length_alleles: repeat length of each allele
 - motif_length: length of the motif
 - bp_length_alleles: bp length of each allele
--- a/str/qc/qc_annotator.py
+++ b/str/qc/qc_annotator.py
@@ -62,16 +62,14 @@ def main(mt_path):
             )
         )
     )
-    # motif length
-    mt = mt.annotate_rows(motif_length=hl.len(mt.info.RU))
 
-    # bp length array annotation: motif length*repeat length
+    # motif length, bp length array annotation: motif length*repeat length
     mt = mt.annotate_rows(
+        motif_length=hl.len(mt.info.RU),
         bp_length_alleles=mt.rep_length_alleles.map(lambda x: x * mt.motif_length)
     )
     # break up allele_1 and allele_2 into separate columns
-    mt = mt.annotate_entries(allele_1_rep_length=hl.int(mt.REPCN.split('/')[0]))
-    mt = mt.annotate_entries(
+    mt = mt.annotate_entries(allele_1_rep_length=hl.int(mt.REPCN.split('/')[0]),
         allele_2_rep_length=hl.if_else(
             hl.len(mt.REPCN.split('/')) == 2,
             hl.int(mt.REPCN.split('/')[1]),
@@ -79,9 +77,7 @@ def main(mt_path):
         )
     )
     mt = mt.annotate_entries(
-        allele_1_bp_length=mt.allele_1_rep_length * mt.motif_length
-    )
-    mt = mt.annotate_entries(
+        allele_1_bp_length=mt.allele_1_rep_length * mt.motif_length,
         allele_2_bp_length=mt.allele_2_rep_length * mt.motif_length
     )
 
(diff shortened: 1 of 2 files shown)


str/qc/qc_annotator.py
```diff
@@ -0,0 +1,223 @@
...
+    mt = mt.annotate_rows(
+        binom_hwep=hl.binom_test(mt.n_hom, mt.n_called, mt.exp_hom_frac, 'two-sided')
+    )
+    mt = mt.drop('n_hom', 'n_called', 'exp_hom_frac')
+
+    # proportion of observed heterozygous calls per locus
+    mt = mt.annotate_rows(obs_het=mt.variant_qc.n_het / mt.variant_qc.n_called)
+
+    # average 'locus coverage' per locus
+    mt = mt.annotate_rows(variant_lc=hl.agg.sum(mt.LC) / mt.variant_qc.n_called)
```
MattWellie, line 165: I don't think there's a functional benefit either way, but you can condense these calls way down by annotating multiple attributes in a single `.annotate_X` call

```
mt = mt.annotate_rows(
    x=y,
    foo=bar
)
```

str/qc/qc_annotator.py
```diff
@@ -0,0 +1,223 @@
...
+
+    # proportion of alleles that are not the mode allele
+    mt = mt.annotate_rows(
+        prop_alleles_is_not_mode=mt.sum_alleles_is_not_mode
+        / hl.sum(mt.aggregated_info.allele_array_counts.values())
+    )
+
+    # add in Gymrek binomial HWEP implementation (Hail Query's HWEP implementation only works for biallelic loci)
+    # https://github.com/gymrek-lab/TRTools/blob/master/trtools/utils/utils.py#L325
+    mt = mt.annotate_rows(n_hom=mt.variant_qc.n_called - mt.variant_qc.n_het)
```
MattWellie, line 152: Hom counts are in the variant_qc struct as `mt.variant_qc.homozygote_count`, an array of ints for `[ref, alt1, [alt2...]]`

str/qc/qc_annotator.py
```diff
@@ -0,0 +1,223 @@
...
+    mt = mt.annotate_rows(exp_hom_frac=hl.sum(mt.variant_qc.AF**2))
+    mt = mt.annotate_rows(n_hom=hl.int32(mt.n_hom))
+    mt = mt.annotate_rows(n_called=hl.int32(mt.variant_qc.n_called))
+    mt = mt.annotate_rows(
+        binom_hwep=hl.binom_test(mt.n_hom, mt.n_called, mt.exp_hom_frac, 'two-sided')
+    )
+    mt = mt.drop('n_hom', 'n_called', 'exp_hom_frac')
```
MattWellie, line 159: if you're immediately dropping these values, why bother moving them? You can run the binomial test on `mt.variant_qc.n_called` etc.

I'd be surprised if variant_qc counts ever returned a non-int value, with a floor of 0

str/qc/qc_annotator.py
```diff
@@ -0,0 +1,223 @@
...
+            hl.missing('int32'),
+        )
+    )
+    mt = mt.annotate_entries(
+        allele_2_REPCI_2=hl.if_else(
+            hl.len(mt.allele_2_REPCI.split('-')) == 2,
+            hl.int(mt.allele_2_REPCI.split('-')[1]),
+            hl.missing('int32'),
+        )
+    )
```
MattWellie, line 193:
```suggestion
    mt = mt.annotate_entries(
        allele_1_REPCI_1=hl.int32(mt.allele_1_REPCI.split('-')[0]),
        allele_1_REPCI_2=hl.int32(mt.allele_1_REPCI.split('-')[1]),
        allele_2_REPCI_1=hl.if_else(
            hl.len(mt.allele_2_REPCI.split('-')) == 2,
            hl.int(mt.allele_2_REPCI.split('-')[0]),
            hl.missing('int32'),
        ),
        allele_2_REPCI_2=hl.if_else(
            hl.len(mt.allele_2_REPCI.split('-')) == 2,
            hl.int(mt.allele_2_REPCI.split('-')[1]),
            hl.missing('int32'),
        )
    )
```

Not 100000% sure I've got the brackets right here, but this can easily be one call

MattWellie (approved):
Lots of this syntax could be condensed down, it'll work, but there's maybe double the number of annotation calls you really need

str/qc/qc_annotator.py
MattWellie: if you want the pure hom counts you can probably do a hl.sum or hl.agg.sum on the array

str/qc/qc_annotator.py
hopedisastro: It needed the casting because I tried first without it and it didn't work :/

PR #163, opened 2024-03-05
Create cell_per_individual_extractor.py

hopedisastro:
This script aims to extract the number of cells per individual per cell type from the anndata object (containing all individuals, all chromosomes).
The output is a CSV file with three columns: CPG ID, cell type, counts of cells.

### DIFF
--- a/str/associatr/helper/cell_per_individual_extractor.py
+++ b/str/associatr/helper/cell_per_individual_extractor.py
@@ -5,7 +5,7 @@ This script aims to extract the number of cells per individual per cell type fro
 The output is a CSV file with three columns: CPG ID, cell type, counts of cells.
 
 analysis-runner --dataset "bioheart" --access-level "test" --description "get cell counts per individual" --output-dir "str/associatr/helper" --image australia-southeast1-docker.pkg.dev/cpg-common/images/scanpy:1.9.3 \
---storage=10G --cpu=4 \
+--storage=10G \
 cell_per_individual_extractor.py --input-h5ad-dir=gs://cpg-bioheart-test/str/anndata/B_naive_chr10.h5ad
 """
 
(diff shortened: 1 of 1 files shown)


str/associatr/helper/cell_per_individual_extractor.py
```diff
@@ -0,0 +1,43 @@
...
+cell_per_individual_extractor.py --input-h5ad-dir=gs://cpg-bioheart-test/str/anndata/B_naive_chr10.h5ad
+"""
+
+import click
+import scanpy as sc
+from cpg_utils.hail_batch import output_path
+from cpg_utils import to_path
+
+
+@click.option('--input-h5ad-dir', help='GCS path to the input anndata object', type=str)
```
MattWellie, line 18:
```suggestion
@click.option('--input-h5ad-dir', help='GCS path to the input anndata object')
```

the default value type is String, so no harm stating it here, but no harm either

str/associatr/helper/cell_per_individual_extractor.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
+
+"""
+This script aims to extract the number of cells per individual per cell type from the anndata object (containing all individuals, all chromosomes).
+The output is a CSV file with three columns: CPG ID, cell type, counts of cells.
+
+analysis-runner --dataset "bioheart" --access-level "test" --description "get cell counts per individual" --output-dir "str/associatr/helper" --image australia-southeast1-docker.pkg.dev/cpg-common/images/scanpy:1.9.3 \
+--storage=10G --cpu=4 \
```
MattWellie, line 8:
```suggestion
--storage=10G \
```
python is single threaded, so probably no benefit here

str/associatr/helper/cell_per_individual_extractor.py
MattWellie: Also `--input-h5ad-dir` as a name is a little misleading, as AFAIK it points to an individual object

str/associatr/helper/cell_per_individual_extractor.py
```diff
@@ -0,0 +1,43 @@
...
+import click
+import scanpy as sc
+from cpg_utils.hail_batch import output_path
+from cpg_utils import to_path
+
+
+@click.option('--input-h5ad-dir', help='GCS path to the input anndata object', type=str)
+@click.option('--version', help='version of the output', default='v1')
+@click.command()
+def main(input_h5ad_dir, version):
```
MattWellie, line 21:
```suggestion
def main(input_h5ad_dir: str, version: str):
```

type hinting is your friend

MattWellie (approved):
It should work ok, but some general suggestions

PR #164, opened 2024-03-13
Runner script for gene-level multiple testing correction 

hopedisastro:
This script computes gene-level p-values using either ACAT/Bonferroni (the first step of multiple testing correction).
Assumed input files follow the format of output TSV files from associaTR.
Output is multiple gene-specific TSV files with the gene name in the first column, and gene-level p-value in the second column.The attributes of the locus with the lowest raw p-value are also stored in the TSV file (coordinates, beta, se, raw pval,r2, motif, ref_len).

Runs ok https://batch.hail.populationgenomics.org.au/batches/437270

### DIFF
--- a/str/associatr/multiple_testing_correction/run_gene_level_pval.py
+++ b/str/associatr/multiple_testing_correction/run_gene_level_pval.py
@@ -108,7 +108,38 @@ def cct(
 
     # write to output
     gcs_output = output_path(
-        f'gene_level_pvals/{cell_type}/chr{chromosome}/{gene_name}_gene_level_pval.tsv',
+        f'gene_level_pvals/acat/{cell_type}/chr{chromosome}/{gene_name}_gene_level_pval.tsv',
+        'analysis',
+    )
+    with to_path(gcs_output).open('w') as f:
+        f.write(
+            f'gene_name\tgene_level_pval\tchr\tpos\tn_samples_tested\tlowest_raw_pval\tcoeff\tse\tr2\tmotif\tref_len\tallele_freq\n'
+        )
+        f.write(
+            f'{gene_name}\t{str(pval)}\t{str(chr)}\t{str(pos)}\t{str(n_samples_tested)}\t{str(raw_pval)}\t{str(coeff)}\t{str(se)}\t{str(r2)}\t{str(motif)}\t{str(ref_len)}\t{str(allele_frequency)}\n'
+        )
+
+
+def bonferroni_compute(
+    gene_name,
+    pvals,
+    cell_type,
+    chromosome,
+    chr,
+    pos,
+    n_samples_tested,
+    raw_pval,
+    coeff,
+    se,
+    r2,
+    motif,
+    ref_len,
+    allele_frequency,
+):
+    pval = min(pvals) * len(pvals)
+    # write to output
+    gcs_output = output_path(
+        f'gene_level_pvals/bonferroni/{cell_type}/chr{chromosome}/{gene_name}_gene_level_pval.tsv',
         'analysis',
     )
     with to_path(gcs_output).open('w') as f:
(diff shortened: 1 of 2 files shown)


str/associatr/multiple_testing_correction/run_gene_level_pval.py
```diff
@@ -0,0 +1,306 @@
...
+                            chr.append(row.iloc[0])
+                            pos.append(row.iloc[1])
+                            n_samples_tested.append(row.iloc[3])
+                            raw_pval.append(row.iloc[5])
+                            coeff.append(row.iloc[6])
+                            se.append(row.iloc[7])
+                            r2.append(row.iloc[8])
+                            motif.append(row.iloc[9])
+                            ref_len.append(row.iloc[11])
+                            allele_frequency.append(row.iloc[12])
```
MattWellie, line 260: Is there a reason for this? Wouldn't iterating over a 1-length object also be valid, so you can loop over the rows even if there's only a single row?

str/associatr/multiple_testing_correction/run_gene_level_pval.py
```diff
@@ -0,0 +1,306 @@
...
+    gcs_output = output_path(
+        f'gene_level_pvals/bonferroni/{cell_type}/chr{chromosome}/{gene_name}_gene_level_pval.tsv',
+        'analysis',
+    )
+    with to_path(gcs_output).open('w') as f:
+        f.write(
+            f'gene_name\tgene_level_pval\tchr\tpos\tn_samples_tested\tlowest_raw_pval\tcoeff\tse\tr2\tmotif\tref_len\tallele_freq\n'
+        )
+        f.write(
+            f'{gene_name}\t{str(pval)}\t{str(chr)}\t{str(pos)}\t{str(n_samples_tested)}\t{str(raw_pval)}\t{str(coeff)}\t{str(se)}\t{str(r2)}\t{str(motif)}\t{str(ref_len)}\t{str(allele_frequency)}\n'
```
MattWellie, line 153: you don't need to wrap these elements in string translations - using an f-string will use either the object's `repr` or `str` methods to get a string representation.

str/associatr/multiple_testing_correction/run_gene_level_pval.py
```diff
@@ -0,0 +1,306 @@
...
+                        chr = [min_rows.iloc[:, 0].values[0]]
+                        pos = [min_rows.iloc[:, 1].values[0]]
+                        n_samples_tested = [min_rows.iloc[:, 3].values[0]]
+                        raw_pval = [min_rows.iloc[:, 5].values[0]]
+                        coeff = [min_rows.iloc[:, 6].values[0]]
+                        se = [min_rows.iloc[:, 7].values[0]]
+                        r2 = [min_rows.iloc[:, 8].values[0]]
+                        motif = [min_rows.iloc[:, 9].values[0]]
+                        ref_len = [min_rows.iloc[:, 11].values[0]]
+                        allele_frequency = [min_rows.iloc[:, 12].values[0]]
```
MattWellie, line 239: In general this is a wild number of values to pass around. I'd personally replace this with something like:

```
# store a mapping of the key description to the index
VALUES_TO_INDEXES = [('chr', 0), ('pos', 1), ...]

# create a dictionary of {key: list}
row_dict = {key: [] for key, value in VALUES_TO_INDEXES}

# populate the dict
for _index, row in min_rows.iterrows():
    for key, value in VALUES_TO_INDEXES:
        row_dict[key].append(row.iloc[value])

# then pass the dict to each method as a single argument

# writing out to a line can be done tidily with 
f.write('\t'.join([dict[key] for key, value in VALUES_TO_INDEXES])
# which would auto-adjust if you add or remove target values/indexes
```

This might not work for your use case. I'm not sure the idiom "If you have a procedure with 10 parameters, you probably missed some" really applies here, but this feels like a lot of separate variables being passed around, especially for the bonferroni method that only changes one of them.

MattWellie (approved):
Not wild about the method interfaces here, this is a very long script passing a lot of separate variables. I suggest changes for code quality/complexity, but functionally probably ok

str/associatr/multiple_testing_correction/run_gene_level_pval.py
hopedisastro: yup good point

PR #165, opened 2024-03-13
Runner script for computing Q-value (FDR correction) 

hopedisastro:
This script performs FDR (across gene) correction (the second and final step of multiple testing correction).
Ensure that `run_acat.py` has been run to generate the gene-level p-values first.
Output is one TSV file per cell type with three columns - gene name, gene-level p-value (ACAT or Bonferroni-corrected), and q-value.

Runs ok https://batch.hail.populationgenomics.org.au/batches/437319


str/associatr/multiple_testing_correction/run_storey.py
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env python3
+# pylint: disable=too-many-arguments,too-many-locals
```
MattWellie, line 2:
```suggestion
```

I don't think these are required...

MattWellie (approved):
It's good practice to include type hints and content in the docstring describing the inputs/data types. It just makes everything easier for when you come back to it in the future

PR #167, opened 2024-03-23
Update get_covariates to select geno PCs 1 and 6 only 

hopedisastro:
See discussion here: https://centrepopgen.slack.com/archives/C063WDT5HT4/p1711155305004049

We amend the script to only genotype PC 1 and 6 as covariates and apply thresholds to these two PCs to remove ancestry outliers; temporary solution until LC pipeline ancestry results are finalised

### DIFF
--- a/str/associatr/get_covariates.py
+++ b/str/associatr/get_covariates.py
@@ -24,11 +24,7 @@ from cpg_utils import to_path
 
 
 def get_covariates(
-    pseudobulk_input_dir,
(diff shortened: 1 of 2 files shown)


MattWellie:
@hopedisastro I've added https://github.com/populationgenomics/sv-workflows/pull/167/commits/fc5a037ec84a0b396db13b629598b03570de8cde to pass linting checks. Not sure if this PR is still required?

PR #175, opened 2024-04-17
Subset scanpy object by sample id list

hopedisastro:
New analysis direction to look at each cohort separately because the WGS batch effects are pretty complex. This PR will subset the andata object to only include cells that have CPG IDs that match those in the file path. 

Scanpy object has data from both cohorts in it by default.


str/associatr/pseudobulk.py
```diff
@@ -25,13 +26,19 @@
...
     """
     Performs pseudobulking in a cell-type chromosome-specific manner
     """
     expression_h5ad_path = to_path(input_file_path).copy('here.h5ad')
     adata = sc.read_h5ad(expression_h5ad_path)
 
+    # retain only samples in the id file
+    with to_path(id_file_path).open() as csvfile:
+        cpg_ids = list(csv.reader(csvfile))
+    cpg_ids = cpg_ids[0]  # the function above creates a list in a list
```
MattWellie, line 39: if this creates a list in a list, why are you wrapping it in list? Isn't that the cause of the double-list?

str/associatr/pseudobulk.py
```diff
@@ -25,13 +26,19 @@
...
-def pseudobulk(input_file_path, target_sum, min_pct):
+def pseudobulk(input_file_path, id_file_path, target_sum, min_pct):
     """
     Performs pseudobulking in a cell-type chromosome-specific manner
     """
     expression_h5ad_path = to_path(input_file_path).copy('here.h5ad')
     adata = sc.read_h5ad(expression_h5ad_path)
 
+    # retain only samples in the id file
+    with to_path(id_file_path).open() as csvfile:
```
MattWellie, line 37: how many IDs are in each cohort? it might be easier to add a list of SG IDs into a config file, and access them that way instead of having to copy a file into GCP then parse it etc.

str/associatr/pseudobulk.py
```diff
@@ -83,6 +90,7 @@ def pseudobulk(input_file_path, target_sum, min_pct):
 
 # inputs:
 @click.option('--input-dir', help='GCS Path to the input AnnData object')
+@click.option('--sample-id-file-path', help='GCS Path to the sample ID file to do the analysis on')
```
MattWellie, line 93: This script is now 9 args, it's worth considering a config file. Relates to other comment on whether adding these SG IDs in config is reasonable

str/associatr/pseudobulk.py
hopedisastro: If I don't wrap it as a list it becomes a CSV reader object, that is not iterable.

str/associatr/pseudobulk.py
hopedisastro: ok i added a config

str/associatr/pseudobulk.py
hopedisastro: It's >900 each time. I prefer to keep it separate as a CSV file so the TOML is cleaner

hopedisastro:
batch run w/ config is good - https://batch.hail.populationgenomics.org.au/batches/443906

str/associatr/pseudobulk.py
MattWellie: Reader objects are designed to be iterable... the following snippet works:
```
with to_path('gs://YOUR_INPUT.csv').open() as h:
    x = csv.reader(h)
    for y in x:
        print(y)
```

`y` here is a 1055 element long list, as your CSV entries are all on one line. If you were trying to access the first element of the results it wouldn't work, as reader objects aren't exactly lists:

```
x = csv.reader(h)
cpg_ids = x[0]
>>> TypeError: '_csv.reader' object is not subscriptable
```

A `csv.reader` object implements a `next` method which allows you to iterate through the underlying file one line at a time for scalability. By wrapping it in `list` you're forcing all those lines to be read at once and stored in memory. Here that's not an issue as all your elements are on one line anyway, but if your CSV was ~GB in size this would be a problem. This is a mini snippet for 'open the file, read the first line':

```
x = csv.reader(to_path('gs://YOUR_INPUT.csv').open())
cpg_ids = next(x)
```

None of this really matters, your implementation works, but it's good to know why it works.

str/associatr/pseudobulk.py
```diff
@@ -82,38 +88,23 @@ def pseudobulk(input_file_path, target_sum, min_pct):
...
     b = get_batch('Run pseudobulk')
+    input_dir = get_config()['pseudobulk']['input_dir']
+    sample_id_file_path = get_config()['pseudobulk']['sample_id_file_path']
+    cell_types = get_config()['pseudobulk']['cell_types']
+    chromosomes = get_config()['pseudobulk']['chromosomes']
+    job_storage = get_config()['pseudobulk']['job_storage']
+    job_memory = get_config()['pseudobulk']['job_memory']
+    job_cpu = get_config()['pseudobulk']['job_cpu']
+    target_sum = get_config()['pseudobulk']['target_sum']
+    min_pct = get_config()['pseudobulk']['min_pct']
```
MattWellie, line 107: This is a nit, but you don't have to (and probably shouldn't) read your config up-front. The config file is universally accessible, so instead of 

```python
job_storage = get_config()['pseudobulk']['job_storage']
job_memory = get_config()['pseudobulk']['job_memory']
job_cpu = get_config()['pseudobulk']['job_cpu']
j = b.new_python_job(name=f'Pseudobulk for {cell_type}: chr{chromosome}')
j.cpu(job_cpu)
j.memory(job_memory)
j.storage(job_storage)
```

It's cleaner to use

```python
j = b.new_python_job(name=f'Pseudobulk for {cell_type}: chr{chromosome}')
j.cpu(get_config()['pseudobulk']['job_cpu'])
j.memory(get_config()['pseudobulk']['job_memory'])
j.storage(get_config()['pseudobulk']['job_storage'])
```

This reduces the number of local variables, and makes it clear to a reader that these values are pulled directly from the config. Instead of reading the script and back-tracking to find out if the storage variable was edited after it was read from config, a user can be assured that the value is exactly as it appears in config. 

This is even more evident when an entry is read from config, then passed as an argument (`sample_id_file_path`, `min_pct`) - I'd just pull this from config during the method for overall clarity

MattWellie (commented):
Code nits, but nothing functional

MattWellie (approved):

PR #177, opened 2024-04-22
Move import cauchy into the function to avoid pickling error 

hopedisastro:
Move the import for the cauchy distribution into the function definition to avoid pickling error. 

https://centrepopgen.slack.com/archives/C030X7WGFCL/p1713737228278469?thread_ts=1713324735.266539&cid=C030X7WGFCL

### DIFF
--- a/str/associatr/multiple_testing_correction/run_gene_level_pval.py
+++ b/str/associatr/multiple_testing_correction/run_gene_level_pval.py
@@ -19,7 +19,7 @@ import pandas as pd
 import hailtop.batch as hb
 
 from cpg_utils import to_path
-from cpg_utils.hail_batch import get_batch, output_path
+from cpg_utils.hail_batch import get_batch, output_path, image_path
 
 # store a mapping of the key description to the index
 VALUES_TO_INDEXES = [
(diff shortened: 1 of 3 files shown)


MattWellie:
can you run `ruff check --fix .` locally to pass the linting checks

str/associatr/multiple_testing_correction/run_gene_level_pval.py
```diff
@@ -65,6 +64,7 @@ def cct(
     #' @export
     R code is implemented in python
     """
+    from scipy.stats import cauchy
```
jmarshall, line 67: It'd be worth putting a comment on this so it doesn't get accidentally “tidied up” in future… something like:

```suggestion
    # Import here as a PythonJob's function must stand alone
    from scipy.stats import cauchy  # noqa: PLC0415
```

(The `noqa` will stop ruff from complaining about this if the relevant check gets applied by default in future.)

jmarshall:
If this is the problem, I wonder if you're going to have the same problem with `to_path` and `output_path`…

hopedisastro:
Matt proposed an optimised runner (#180) because the current implementation was very inefficient in getting the driver job to complete when running using the full dataset using all cell types

MattWellie:
Apologies, the diff now looks awful 😓 

- can you reduce the concurrency permitted to remove those errors you saw in test
- can you add an escape for 'this file already exists' to get rid of those preemption problems (concurrency cap might solve that anyway)

Happy to have a crack at those, or if you're happy to just forge ahead with the concurrency change

MattWellie (approved):
approved on the basis that the concurrency CLI argument will be reduced when used

str/associatr/multiple_testing_correction/run_gene_level_pval.py
```diff
@@ -163,16 +215,8 @@ def bonferroni_compute(
     default=500,
```
hopedisastro, line 215:
```suggestion
    default=18,
```

PR #178, opened 2024-04-22
Extracts raw p-values for association results by cell type into a text file 

hopedisastro:
This script extracts the raw p-values from the results of associaTR into one text file per cell type.
For downstream use to make a QQ plot.


michael-harper (approved):
LGTM

PR #181, opened 2024-04-22
QQPlotter

hopedisastro:
This script plots a QQ plot of observed vs expected -log10(p-values) for each cell type and saves the plot in analysis bucket


hopedisastro:
FYI it says 4 files changed but 3 files are just to pass linting checks. Only real code change is the qqplotter.py

michael-harper (approved):

PR #182, opened 2024-04-25
Meta analysis runner

hopedisastro:
This script runs R's meta package to generate pooled effect sizes, and p-values for each eQTL.
Outputs a TSV file per gene to gcp. 

I will update the image path with the correct image prior to merging into `main`, once the [image's PR](https://github.com/populationgenomics/images/pull/141) is approved


MattWellie:
I don't really like the design of this PR as it stands... It may work, but I think it's bad practice:

1. The process we're trying to run is a R module
2. We're running that R module in python using rpy2
3. We're also making the job read from and write directly to GCP, which requires GCloud authentication libraries to be installed
4. Because we're requiring all these behaviours, we're using a huge docker image (~6GB)

I would advise re-thinking this PR and the corresponding image:

1. Running using rpy2/`python_job` is fine, it makes running the R code much easier... However, that image should only really need Python, rpy2, pandas, and the R installation/packages
2. The job itself doesn't need to read/write from GCP, reading in/out can be delegated to hail batch
3. If we delegate file I/O to Hail Batch, we can remove a ton of stuff from the docker image

MattWellie (approved):

PR #183, opened 2024-04-27
ACAT gene level pvalue script - but modified to accept the input file from the meta-analysis output

hopedisastro:
Same logic as https://github.com/populationgenomics/sv-workflows/blob/main/str/associatr/multiple_testing_correction/run_gene_level_pval.py, but just amended so that it expects an input file of the format that is outputted from meta-analysis (meta_runner.py).

### DIFF
--- a/str/associatr/meta_analysis/run_gene_level_pvals_meta.py
+++ b/str/associatr/meta_analysis/run_gene_level_pvals_meta.py
@@ -174,7 +174,7 @@ def cct(gene_files: list[str], cell_type: str, chromosome: str, og_weights=None)
         )
         with to_path(gcs_output).open('w') as f:
             f.write(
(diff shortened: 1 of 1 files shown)


str/associatr/meta_analysis/run_gene_level_pvals_meta.py
```diff
@@ -0,0 +1,276 @@
...
+    #' @export
+    R code is implemented in python
+    """
+    # Import here as a PythonJob's function must stand alone
+    from scipy.stats import cauchy  # noqa: PLC0415
+
+    from cpg_utils import to_path
+    from cpg_utils.config import output_path
+
+    for gene_file in gene_files:
```
MattWellie, line 123: possible check here - if output file already exists, skip onto the next gene

This has happened before when the jobs are preempted/restarted, and it starts from the beginning of the gene list again. Running with full/test permissions will mean that it doesn't fail, but avoiding the repeated loops should mean you can run using standard permissions

MattWellie (approved):
LGTM

PR #184, opened 2024-04-27
Add gene name to the lines in the output file 

hopedisastro:


michael-harper (approved):
Input directory living in the analysis bucket?

PR #185, opened 2024-04-30
QC script applying filters to produce chr-specific VCF files for SAIGE-QTL

htanudisastro:
This script applies filters to a ExpansionHunter MT, and outputs chromosome-specific VCF ready for input into SAIGE-QTL.
Input MT should be the output of qc_annotator.py

Applied filters:
- remove monomorphic variants
- locus-level callrate >=0.9
- sample-level call rate >=0.99
- observed heterozygosity >=0.00995 (corresponds to a MAF = 0.5%)
- locus-level HWEP (binom definition) >=10^-6
- set calls outside of [-30,20] relative to mode to NA
- enforce locus-level call rate >=0.9
- set NA alleles to mode allele for the respective locus (SAIGE-QTL does not support missing alleles)

### DIFF
--- a/str/qc/qc_filters_saige.py
+++ b/str/qc/qc_filters_saige.py
@@ -1,0 +1,229 @@
+#!/usr/bin/env python3
+# pylint: disable=missing-function-docstring,no-member,unnecessary-lambda
+"""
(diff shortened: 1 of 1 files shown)


hopedisastro:
successful batch - https://batch.hail.populationgenomics.org.au/batches/444895

PR #189, opened 2024-05-03
Calculate pairwise LD between an STR locus and all SNPs in a specified window

hopedisastro:
This script calculates pairwise LD (correlation coefficient) between one STR locus and all SNPs in a specified window. Helper script for coloc analysis. 

Outputs locus-level LD results as a CSV file. 

Test batch run https://batch.hail.populationgenomics.org.au/batches/445118


michael-harper (approved):
Not sure how large the vcf's files are but could this be made quicker using bcftools to parse vcf files?

PR #193, opened 2024-05-08
Calculate LD between one STR locus and all SNPs in a colocalised locus. 

hopedisastro:
Step after colocalisation. 

Essentially, colocalisation compares signals between SNP eQTLs and SNPs in a GWAS (whether the signals look similar at a gene expression and a phenotype level respectively). 

Most GWAS catalogs are SNPs. So when we add STRs into the mix we do the usual SNP-SNP colocalisation between the eQTLs and GWAS, and then we can link in the STR by showing that the top eSTR associated with that gene is in LD with at least one SNP in the GWAS locus. 

This script calculates the LD. 

Specifically, the workflow: 
1)Extract genes that have >50% posterior probability of a shared causal variant from coloc results.
2) Extract the coordinates of the cis-window (gene +/- 100kB) for each gene using a gene annotation file.
3) Extract the top STR locus (ie passed FDR 5% threshold) for each gene in 1). May be multiple STRs per gene (if tied for smallest ACAT-corrected p-value).
4) Run LD parser() which:
- Extracts GTs for all SNPs (from a chr-specific VCF) in the cis-window for a gene.
- Extracts GTs for the specified STR locus associated with the gene.
- Calculates pairwise correlation of every SNP locus with the target STR locus.
- Save the SNP with the highest absolute correlation to a TSV file. Output to GCP

### DIFF
--- a/str/coloc/ld_parser.py
+++ b/str/coloc/ld_parser.py
@@ -1,18 +1,28 @@
 #!/usr/bin/env python3
 
 """
-This script calculates pairwise LD (correlation coefficient) between one STR locus and all SNPs in a specified window.
 
-Outputs the locus-level LD results as a TSV file.
+Assumes coloc_runner.py and coloc_results_parser.py have been run previously.
+
+This script calculates pairwise LD (correlation coefficient) between one STR locus and all SNPs in a colocalised locus.
+
+Workflow:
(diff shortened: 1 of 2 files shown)


michael-harper (approved):
Looks good!

PR #196, opened 2024-05-09
Parse/combine the results of coloc_runner.py

hopedisastro:
This script concatenates the results of running `coloc_runner.py` (output is per gene) into a single CSV file.

Fine to write to analysis bucket because this is gene level association information (not individual identifying)


hopedisastro:
batch: https://batch.hail.populationgenomics.org.au/batches/445369

michael-harper (approved):

PR #198, opened 2024-05-09
add cpu and storage param 

hopedisastro:
`Main` run ran out of storage space. Just adding params to the job so it is more flexible. https://batch.hail.populationgenomics.org.au/batches/445375


MattWellie (approved):
It would be nice to move some of these into a config file, 9 CLI arguments is a lot (not a priority)

PR #200, opened 2024-05-13
Coloc runner for Gymrek UKBB STR GWAS catalogs 

hopedisastro:
This script performs colocalisation analysis betweeen eGenes identified by pseudobulk STR analysis and GWAS signals from UKBB Gymrek STR catalogs.
1) Identify eGenes (FDR <5%) from the STR analysis
2) Extract the STR GWAS data for the cis-window (gene +/- 100kB)
3) Run coloc for each eGene (if STR eQTL data is available)
4) Write the results to a TSV file

### DIFF
--- a/str/coloc/coloc_gymrek_ukbb_runner.py
+++ b/str/coloc/coloc_gymrek_ukbb_runner.py
@@ -19,6 +19,7 @@ import gzip
 import click
 import pandas as pd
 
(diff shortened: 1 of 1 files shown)


michael-harper (approved):
I can't speak much to the R script being run inside rpy2, but the flow looks good

PR #201, opened 2024-05-13
STR - GWAS lead SNP LD calculator 

hopedisastro:
This script calculates pairwise LD (correlation coefficient) between the top eSTR and the lead SNP (per gene) in the GWAS.
(DOES NOT REQUIRE coloc_runner.py and coloc_results_parser.py to be run previously, nor does it run `coloc`.)

Workflow:
1) Extract genes that are associated with the eSTRs (FDR <5%)
2) Extract the SNP GWAS data for the cis-window (gene +/- 100kB)
3) Select the lead SNP from 2) (ie SNP with the lowest p-value)
4) Calculate pairwise correlation of the genotypes of the lead eSTR locus with those of the lead SNP using cohort-specific VCFs 
Outputs this correlation as a TSV file into `analysis` bucket. 

Successful test batch - https://batch.hail.populationgenomics.org.au/batches/446151

Heavily adapted from the `ld_runner.py` in main (now renamed to `coloc_ld_runner.py` because both scripts calculate LD between SNP genotypes at certain loci with the target STR. Main difference is that this script doesn't parse the coloc results whereas the `ld_runner.py` (now called `coloc_ld_runner.py`) integrates the coloc results for filtering purposes prior to calculating LD.


MattWellie (approved):
LGTM! (But I think I had some design influence, so I would say that...)

PR #202, opened 2024-05-13
Add concurrency cap to gwas ld parser script. 

hopedisastro:
It'll be a lot of eGenes...


str/coloc/gwas_ld_runner.py
```diff
@@ -127,6 +128,7 @@ def ld_parser(
 )
 @click.option('--job-cpu', default=1)
 @click.option('--job-storage', default='20G')
+@click.option('--max-parallel-jobs', default=500)
```
MattWellie, line 131: can Hail take 500 jobs running at once?

PR #203, opened 2024-05-14
Speed up gwas ld

hopedisastro:
Current implentation takes a looonnng time for driver job to kick off https://batch.hail.populationgenomics.org.au/batches/447610/jobs/1

Moved some parts into the helper function so hopefully it speeds things up 

Runs ok in test https://batch.hail.populationgenomics.org.au/batches/447650

### DIFF
--- a/str/coloc/gwas_ld_runner.py
+++ b/str/coloc/gwas_ld_runner.py
@@ -20,7 +20,7 @@ analysis-runner --dataset "bioheart" \
     --str-vcf-dir=gs://cpg-bioheart-test/str/saige-qtl/input_files/vcf/v1-chr-specific \
     --gwas-file=gs://cpg-bioheart-test/str/gwas_catalog/hg38.EUR.IBD.gwas_info03_filtered.assoc_for_gwas_ld.csv \
     --celltypes=CD4_TCM \
-    --phenotype=ibd
+    --phenotype=ibd \
     --max-parallel-jobs 250
 
 """
@@ -186,9 +186,8 @@ def main(
         str_fdr = str_fdr[str_fdr['qval'] < 0.05]  # subset to eGenes passing FDR 5% threshold
 
         # obtain inputs for LD parsing for each entry in `str_fdr`:
-        #for index, row in str_fdr.iterrows():
-        for gene in ['ENSG00000277301']:
-            #gene = row['gene_name']
+        for index, row in str_fdr.iterrows():
+            gene = row['gene_name']
 
             # obtain snp cis-window coordinates for the gene
             gene_table = gene_annotation_table[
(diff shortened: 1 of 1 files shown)


hopedisastro:
TBH I don't really know why it is taking so long to kick start the driver job. I'm hoping that these changes fix the issue, but won't really be able to tell until we run in `main`.

MattWellie:
I've got a feeling that these huge startup times are caused by hail's job dependencies. If your batch has 100000 jobs and a max running of 250, you have (100000-250) jobs with dependencies, (1000000-500) of those have a dependency on a dependent job... and it's close to exponential. 

You can smash this up using a new batch per... cell type? Chromosome? Using the reset batch functionality in between to spin off a new batch. I'd be tempted to try that; in test you never have the numbers to hit this issue

hopedisastro:
I just moved everything to the helper function. I think this should work? @MattWellie can we give this a spin, or should I also incorporate reset_batch, but still wary of the empty deque error.

MattWellie:
> still wary of the empty deque error.

This is the script at the state which was causing that dequeue error: [here](https://github.com/populationgenomics/sv-workflows/blob/7a9a788aee5d2b34f156f79c9d5e730e931f5754/str/coloc/coloc_gymrek_ukbb_runner.py)

I think there's a bug here - the dependency management is done across all batches, but the batches are reset and started with each run. That means that once the job counts get high enough a batch will have dependencies against completely different batches, which would cause an error. The dependency list should also be reset each time the batch is reset

str/coloc/gwas_ld_runner.py
```diff
@@ -30,74 +31,150 @@
...
+        end = str(int(pos) + 1)
+        str_locus = f'{chr_num}:{pos}-{end}'
+        write_path = output_path(
+            f'gwas_ld/{phenotype}/{celltype}/{gene}_chr{chr_num}_{pos}_gwas_ld_results.csv',
+            'analysis',
+        )
+
+        if to_path(write_path).exists():
+            print(f'GWAS LD for {gene} and {str_locus} already exists. Skipping...')
+            continue
```
MattWellie, line 103: suggestion here - if you do a `listdir` operation on the whole directory, you won't need a separate, relatively expensive exists check on each individual file

```
output_path(
            f'gwas_ld/{phenotype}/{celltype}/*.csv',
            'analysis',
        )
```

str/coloc/gwas_ld_runner.py
```diff
@@ -30,74 +31,150 @@
...
+
+        print(f'Running LD for {gene} and {str_locus}')
+        snp_vcf_path = f'{snp_vcf_dir}/chr{chr_num}_common_variants.vcf.bgz'
+        str_vcf_path = f'{str_vcf_dir}/hail_filtered_chr{chr_num}.vcf.bgz'
```
MattWellie, line 107: the nesting here means that you re-copy this per-chromosome VCF for every gene? Instead of copying to a generic 'local' name, can you copy it to 'local_{chr}_STR/SNP.vcf', and only copy a new one in if you haven't already copied that down? 

I may be misunderstanding the looping, but I'm assuming that the same chromosome can appear multiple times on line 92

str/coloc/gwas_ld_runner.py
MattWellie: This is not a priority, but individual existence checks can really extend runtimes

MattWellie (approved):

str/coloc/gwas_ld_runner.py
hopedisastro: moved the local paths before the for loop

PR #204, opened 2024-05-15
GWAS-LD script update (I hope this is the last one)

hopedisastro:
The latest run was 1) a bit expensive and 2) cascade of cancelled jobs failing because they were 'dependent' on a failed job
https://batch.hail.populationgenomics.org.au/batches/447694

This update: 
1) Batches jobs by chromosome (not gene). This means that there are 22 jobs per cell type. This should reduce the cost because big files (eg GWAS catalog and the gene annotation files) are only read once per chromosome (instead of being re-read for every gene). This should also fix the cancelled jobs issue because we can take out the max-concurrent cap which creates dependency between jobs that are actually independent from each other. 

2) Revert back to reading in SNP VCF and STR VCF using batch temp dir - hopefully this solves the issue of batch failing to use the mounted storage.

### DIFF
--- a/str/coloc/gwas_ld_runner.py
+++ b/str/coloc/gwas_ld_runner.py
@@ -13,10 +13,10 @@ Workflow:
 
 analysis-runner --dataset "bioheart" \
     --description "Calculate LD between STR and SNPs" \
-    --access-level "full" \
+    --access-level "test" \
     --cpu=1 \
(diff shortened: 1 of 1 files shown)


hopedisastro:
test batch is ok https://batch.hail.populationgenomics.org.au/batches/447702

MattWellie (approved):
Looks like this should be much more efficient

PR #206, opened 2024-05-15
Add snp_in_vcf flag to fix logic error. 

hopedisastro:
Batch runs, faster and cheaper. 

But logic error - https://batch.hail.populationgenomics.org.au/batches/447720/jobs/11

I think having a flag that signifies whether the target SNP is in the VCF should fix the error.


MattWellie (approved):
I'd usually go True/False, and `if not snp_in_vcf_flag: ...`, but this works

PR #207, opened 2024-05-16
Another logic error

hopedisastro:
Got another logic error where a variable was referenced before assignment at line 144. 

While I'm at it, I also just added 1) chromosome dynamic param and 2) manage concurrency (if ever needed).


michael-harper (approved):
Looks good

PR #210, opened 2024-05-17
Create a SNP VCF in the style of ExpansionHunter VCF for direct use with associaTR

hopedisastro:
This allows us to run the pseudobulk associaTR pipeline as is, just swapping the STR VCF for the SNP VCF because this script amend the format of the SNP VCF to look like ExpansionHunter. 

Briefly: 
- add in EH specific FORMAT header lines 
- remove 'CPG' prefix from sample IDs because associatr requires sample ids to be strictly numeric 
- add in constant value of 3 for 'REF' and 'RU' fields because they need to take a positive whole number value 
-  add in constant 0 for 'RL' - this is so that the REF allele will automatically default to 0 
- update ALT column to always store <\STR1\>, consistent with EH formatting; ensures that the ALT allele is given a value of 1.

### DIFF
--- a/str/associatr/helper/snp_vcf_for_associatr.py
+++ b/str/associatr/helper/snp_vcf_for_associatr.py
@@ -98,7 +98,7 @@ def reformat_vcf(vcf_file_path, output_file_path):
 def main(vcf_dir, chromosomes, job_storage, job_cpu):
     b = get_batch()
     for chrom in chromosomes.split(','):
-        snp_vcf = f'{vcf_dir}/chr{chrom}_common_variants.vcf.gz'
+        snp_vcf = f'{vcf_dir}/chr{chrom}_common_variants.vcf.bgz'
         output_file = output_path(f'common_variants_snps/hail_filtered_chr{chrom}.vcf')
         reformatting_job = b.new_python_job(name=f'Reformating chr{chrom} VCF')
         reformatting_job.storage(job_storage)
--- a/str/associatr/helper/snp_vcf_for_associatr.py
+++ b/str/associatr/helper/snp_vcf_for_associatr.py
@@ -1,4 +1,4 @@
-##!/usr/bin/env python3
+#!/usr/bin/env python3
 
 """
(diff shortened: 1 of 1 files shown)


hopedisastro:
I've tested this in a jupyter notebook and this newly formatted SNP VCF runs with associaTR

str/associatr/helper/snp_vcf_for_associatr.py
```diff
@@ -0,0 +1,112 @@
...
+                # Process variant lines
+                parts = line.strip().split('\t')
+                parts[0] = 'chr' + parts[0]
+                info_field = parts[7]
+                format_field = parts[8]
+                sample_data = parts[9:]
+
+                # Extract PID from the sample data if available
+                pid_index = format_field.split(':').index('PID') if 'PID' in format_field.split(':') else None
+                pid = sample_data[0].split(':')[pid_index] if pid_index is not None else '.'
```
michael-harper, line 63: Why would PID be unavailable?

michael-harper (approved):
Pretty straight forward editing of lines in a VCF, nice job 👍
Just clarifying that format field with have majority missing values except for GT?
`GT:ADFL:ADIR:ADSP:LC:REPCI:REPCN:SO:QUAL` with an example of the sample data being `0/1:.:.:.:.:.:.:.:.`

str/associatr/helper/snp_vcf_for_associatr.py
hopedisastro: Yup; 

> Just clarifying that format field with have majority missing values except for GT?
> GT:ADFL:ADIR:ADSP:LC:REPCI:REPCN:SO:QUAL with an example of the sample data being 0/1:.:.:.:.:.:.:.:.
associaTR only pulls out the values from GT; so we can just leave the others blank rather than put in decoy values . 

>Why would PID be unavailable?
I don't know either, in the SNP VCF I see some with PIDs (like a unique string identifier that describes the locus), and others without.

PR #211, opened 2024-05-17
Bgzip and tabix helper file 

hopedisastro:
This script receives a VCF file and performs BGZIP and TABIX
Successful test batch - https://batch.hail.populationgenomics.org.au/batches/448754

### DIFF
--- a/str/helper/bgzip_tabix.py
+++ b/str/helper/bgzip_tabix.py
@@ -5,7 +5,7 @@ This script receives a gzipped VCF file and performs BGZIP, writing files to out
 
 analysis-runner --access-level test --dataset bioheart --description  \
     'VCF combiner'  --output-dir 'str/associatr/common_variant_snps' \
-    bzip.py \
+    bgzip_tabix.py \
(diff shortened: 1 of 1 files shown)


michael-harper (approved):
Looks good. Just wanted to clarify that these log outputs wont cause issues downstream? Specifically the `PL` and `chr20` definitions in the headers?
From the batch logs:
```
[W::bcf_hdr_check_sanity] PL should be declared as Number=G
[W::vcf_parse] Contig 'chr20' is not defined in the header. (Quick workaround: index the file with tabix.)
[W::vcf_parse_info] INFO '.' is not defined in the header, assuming Type=String
[W::bcf_hrec_check] Invalid tag name: "."
```

hopedisastro:
I've fixed this error: 
[W::vcf_parse_info] INFO '.' is not defined in the header, assuming Type=String
[W::bcf_hrec_check] Invalid tag name: "."

Confirming that the other two warnings preceding it didn't seem to have an effect on `test`

PR #212, opened 2024-05-17
Skips if output file already exists; add always_run()

hopedisastro:
Am expecting some genes to fail. But I don't want a cascade of cancelled jobs so adding always_run()


MattWellie (approved):

michael-harper (approved):
Job inside the batch is not cancelable but the Batch itself is 👌

PR #216, opened 2024-05-20
Add ALT allele to the "RU" field in the mock EH VCF

hopedisastro:
The current state of the script implements a dummy '3' value for "RU" for all loci. 

Here I propose adding the REF/ALT allele in into the "RU" field. 

This allows us to retain the REF/ALT info in the association output files, which is crucial where there are multiple variants with the same CHR:POS coordinates (eg multi allelic loci). 

Have tested this and runs as expected.

### DIFF
--- a/str/associatr/helper/snp_vcf_for_associatr.py
+++ b/str/associatr/helper/snp_vcf_for_associatr.py
@@ -66,7 +66,14 @@ def reformat_vcf(vcf_file_path, output_file_path):
 
                 # Update INFO field (add new fields required by ExpansionHunter, filler fields eg REF =0)
                 # RU stores the alt_allele; this will allow us to discriminate between loci with the same POS
-                new_info_fields = [f'END={parts[1]}', 'REF=3', f'REPID={pid}', 'RL=0', f'RU={parts[3]}-{parts[4]}', f'VARID={pid}']
+                new_info_fields = [
+                    f'END={parts[1]}',
+                    'REF=3',
+                    f'REPID={pid}',
+                    'RL=0',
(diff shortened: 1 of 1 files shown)


str/associatr/helper/snp_vcf_for_associatr.py
```diff
@@ -64,8 +64,16 @@ def reformat_vcf(vcf_file_path, output_file_path):
...
-                # Update INFO field (add new fields required by ExpansionHunter, filler fields eg REF and RU are a constant 3)
-                new_info_fields = [f'END={parts[1]}', 'REF=3', f'REPID={pid}', 'RL=0', 'RU=3', f'VARID={pid}']
+                # Update INFO field (add new fields required by ExpansionHunter, filler fields eg REF =0)
+                # RU stores the ref and alt_allele; this will allow us to discriminate between loci with the same POS
+                new_info_fields = [
+                    f'END={parts[1]}',
+                    'REF=3',
+                    f'REPID={pid}',
+                    'RL=0',
+                    f'RU={parts[3]}-{parts[4]}',
```
michael-harper, line 74: Worried about populating a field with misleading information and not having any way of going back and deciphering.
Can we change the header information on lines #L44 to something like this?
```
fout.write('##INFO=<ID=RU,Number=1,Type=String,Description="Storing the REF/ALT info instead of Repeat Unit to retain the REF/ALT info in the association output files, which is crucial where there are multiple variants with the same CHR:POS coordinates (eg multi allelic loci)">\n')

michael-harper (approved):
Have approved if you think my suggestion is worth including

PR #219, opened 2024-05-21
"Meta sandwich" or DF concatenator. 

hopedisastro:
This script concatenates (row-wise) two dataframes together.
Assumed that the header for the two dataframes are the same.
Outputs results as a TSV file.

This helper script will be used to concatenate the results of the meta-analysis for eSNPs and eSTRs together, with each gene having its own file.
Only common genes between the two datasets will be concatenated.

Batch runs: https://batch.hail.populationgenomics.org.au/batches/453262

### DIFF
--- a/str/associatr/meta_analysis/dataframe_concatenator.py
+++ b/str/associatr/meta_analysis/dataframe_concatenator.py
@@ -24,7 +24,7 @@ import click
 import hailtop.batch as hb
 
 from cpg_utils import to_path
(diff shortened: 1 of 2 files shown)


michael-harper (approved):
Do you need to check if the file exists in `input_dir1`?

hopedisastro:
no, it uses input_dir_1 as the reference base and compares if there's a matching file in input_dir_2 for every file in dir 1

PR #220, opened 2024-05-21
Use a file-of-filenames for mergeSTR input file arguments

jmarshall:
As a followup to PR #151 et al, for when we have TRTools v6.x in an image for use by this script, here is a suggested way to change it to use `--vcfs-list` instead of its current hacks to squeeze all the filenames onto the command line.

Since TRTools v6.0, mergeSTR has a `--vcfs-list` option that takes a list file containing filenames as an alternative to listing them on the command line. Use this to entirely avoid line length limits when merging thousands of VCFs.

Write the `cat ... EOF` command all on one source line, so that there is no indentation on the input data and `EOF` line in the shell snippet, which otherwise wouldn't work. (Eventually we may have hail-is/hail#14356 or similar to facilitate getting the list of filenames to the worker, but at the moment this `cat` is probably the least horrible way.)

(Uses a newline helper variable, as Python 3.10 f-strings can't contain backslashes such as `f"""...{"\n"...}..."""`.)

### DIFF
--- a/str/trtools/merge_str_runner.py
+++ b/str/trtools/merge_str_runner.py
@@ -90,11 +90,15 @@ def main(
         if len(cpg_ids) != len(set(cpg_ids)):
             raise ValueError('Duplicate CPG IDs detected in sample list')
 
-        # TODO Convert to use mergeSTR --vcfs-list when that option is available
+        batch_vcfs_list = '${BATCH_TMPDIR}/batch_vcfs.list'
+
+        newline = '\n'
+        # Writing cat...EOF on one line avoids any indentation on the batch_vcfs and EOF lines
         trtools_job.command(
             f"""
-        cd ${{BATCH_TMPDIR}}/inputs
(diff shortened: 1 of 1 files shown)


hopedisastro (approved):

MattWellie:
An alternative here would be a find command. As the VM starts, run a find from ${BATCH_TMPDIR} for all VCFs. This would dodge any issues from needing exact paths due to the hashes in paths.

MattWellie:
An alternative here would be a find command. As the VM starts, run a find from ${BATCH_TMPDIR} for all VCFs. This would dodge any issues from needing exact paths due to the hashes in paths.

jmarshall:
That would be an alternative, but IMHO as the script already has the exact set of desired VCF paths in `batch_vcfs` it's better to use them.

Finding all VCFs under $BATCH_TMPDIR or `…/inputs` is fuzzier, e.g., if the script later gains another VCF input file that it uses for some other purpose, `find` would pick it up too.

PR #222, opened 2024-05-23
Condition on the SNP genotype for eQTL analysis 

hopedisastro:
Further workup of results require conditioning on the lead SNP genotype to see if the signal persists after we take out the SNP. 
Practically this means adding the SNP genotype for the lead SNP locus as a new covariate

This PR implements that change: 
- user can specify one or multiple target SNP loci to control for as `snp_loci` in the config toml. 
- helper method `extract_genotypes` extracts the GT for the relevant locus (0,1,2 are the only GT options) 
- This `sno_genotype_vcf` is then merged with the larger `gene_pheno_cov` (which is a larger df containing other existing covariates), based on the 'sample_id'

### DIFF
--- a/str/associatr/get_cis_numpy_files.py
+++ b/str/associatr/get_cis_numpy_files.py
@@ -203,7 +203,7 @@ def main():
                 snp_vcf_dir = get_config()['get_cis_numpy']['snp_vcf_dir']
                 snp_vcf_path = f'{snp_vcf_dir}/{chrom}_common_variants.vcf.bgz'
                 snp_input = get_batch().read_input_group(**{'vcf': snp_vcf_path, 'csi': snp_vcf_path + '.csi'})
-            else: # no SNP VCF provided
+            else:  # no SNP VCF provided
                 snp_input = None
             j.call(
                 cis_window_numpy_extractor,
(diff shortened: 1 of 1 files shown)


hopedisastro:
Works in test batch and I have also printed a test CSV where I confirm the covariate df looks the way iI want it to. https://batch.hail.populationgenomics.org.au/batches/453857

str/associatr/get_cis_numpy_files.py
```diff
@@ -31,6 +32,45 @@
...
+                for idx, sample in enumerate(record.genotypes):
+                    genotype_call = sample[0] + sample[1]
+                    if genotype_call == 0:
+                        genotype = "0"  # HOM REF
+                    elif genotype_call == 1:
+                        genotype = "1"  # HET
+                    elif genotype_call == 2:
+                        genotype = "2"  # HOM ALT
+                    else:
+                        genotype = "."
```
MattWellie, line 66: The cyvcf2 representation already has a cleanly parsed way of getting this information 
 as np arrays: https://github.com/brentp/cyvcf2?tab=readme-ov-file#example
```
[variant.]gt_types is array of 0,1,2,3==HOM_REF, HET, UNKNOWN, HOM_ALT
```

MattWellie (approved):
Happy if you're happy, just flagging that the GT representations aren't equivalent between what you had before and what cyvcf provides
[ref](https://github.com/populationgenomics/sv-workflows/pull/222/files/6de137242596579a592b1b5303dfe8800bd0b95a..af6bdf7afab57f869870c7305716750050b6f5c4)

PR #227, opened 2024-06-10
Filter out indels that are actually STRs, as well as duplicate STRs. 

hopedisastro:
This script filters raw associaTR outputs that contain both eSTR and eSNP results (i.e. we assume that `dataframe_concatenator.py` has been run).
Particularly, we remove indels that actually represent STRs, AND remove duplicate eSTRs (retain only one eSTR per duplicate set).
This is necessary to improve the accuracy of fine-mapping.
Indels are considered STRs (and subsequently removed) if:
- they overlap with an STR region, specified by an eSTR in the associaTR output; and
- the (reverse complement) sequence of the indel is a whole/partial copy of at least one cyclical representation of the STR motif.
e.g. a 'TAT' insertion overlapping with a 'ATT' STR would be considered an STR as 'TAT' is a cyclical representation of 'ATT'.
Similarly, a 'GC' insertion overlapping with a 'CAG' STR would be considered an STR, as 'GC' is a partial copy of 'GCA', a cyclical representation of 'CAG'.
Note that impure indels are conservatively not considered STRs. For example, a 'GCCGCA' insertion overlapping a 'GCC' STR would not be considered an STR.

This script additionally removes duplicate eSTRs (defined by sharing the same coordinates and motif), retaining only one eSTR per duplicate set (chosen based on having the lowest p-value).


str/associatr/fine-mapping/remove_STR_indels.py
```diff
@@ -0,0 +1,234 @@
...
+                            found = True
+                            break
+                        # print(f"Indel: {row1['motif']} doesn't match {str_motif}, position: {row2['pos']}")
+                    found = True
+                    break
+            # if not found:
+            # print("Indel not in STR interval")
+        # Drop the rows where indels are representing STRs
+        result_df = result_df.drop(indices_to_delete)
+        result_df.to_csv(output_path(f'{celltype}/{chrom}/{gene_file_name}'), sep='\t', index=False)
```
hopedisastro, line 183:
```suggestion
        result_df.to_csv(output_path(f'{celltype}/{chrom}/{gene_file_name}', 'analysis'), sep='\t', index=False)
```

silkm (approved):
Hi Hope, I've gone over this as best I can and all looks good.

PR #229, opened 2024-06-12
use EH-style SNP VCFs as input

hopedisastro:
This change uses the EH-style SNP VCFs as input to calculate the LD matrix, not the output vcfs from SAIGE. 

LD matrices are the only output, and outputted to the 'analysis' bucket - no security risk because it's not individual level information.


michael-harper (approved):
LGTM

PR #231, opened 2024-06-14
Finemap runner

htanudisastro:
Runner script to run FINEMAP (a finemapping tool)


str/fine-mapping/finemap_runner.py
```diff
@@ -0,0 +1,36 @@
...
+    eh_job.command(
+                    f"""
+                # Create a temporary file
+                temp_file=$(mktemp)
+
+                input_dir=$(ls -d ${{BATCH_TMPDIR}}/inputs/*/)
+
+                # Write the required format to the file
+                echo 'z;ld;snp;config;cred;log;k;n_samples' > $temp_file
+                echo '${{input_dir}}data.z;${{input_dir}}data.ld;${{input_dir}}data.snp;data.config;data.cred;data.log;/data.k;5363' >> $temp_file
```
jmarshall, line 32: Creating the master file at runtime like this via `echo` is the right thing to do. To get the input group filenames into it you'd want to do something like
```suggestion
                echo "{data_in.z};{data_in.z};{data_in.ld};…etc…;5363" >> $temp_file
```

hopedisastro: ah okay, thanks! I was messing with batch_tmpdir, but this makes sense and is much cleaner :)

jmarshall:
Probably also going to be really useful to add this while you're debugging :smile:

```
cat $temp_file
```

str/fine-mapping/finemap_runner.py
jmarshall: Double quotes `"` (rather than single `'` quotes)  is going to be important here BTW because the `{data_in.foo}` Hail things will expand to `$BATCH_TMPDIR}/blah/blah` and that variable will need to be expanded by the shell before finemap sees it.

str/fine-mapping/finemap_runner.py
```diff
@@ -0,0 +1,177 @@
+#!/usr/bin/env python3
+"""
+This script runs FINEMAP.
+
+analysis-runner --dataset "bioheart" --access-level "test" --description "Run finemap" --output-dir "str/associatr/fine_mapping" \
+     --image "australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:d4922e3062565ff160ac2ed62dcdf2fba576b75a-hail-8f6797b033d2e102575c40166cf0c977e91f834e" \
```
MattWellie, line 6: afaik this is the default driver image, so you'll get this without specifying an image. Unless this points to a specific non-current image?

str/fine-mapping/finemap_runner.py
```diff
@@ -0,0 +1,177 @@
...
+        """
+        To avoid having too many jobs running at once, we have to limit concurrency.
+        """
+        if len(_dependent_jobs) >= max_parallel_jobs:
+            job.depends_on(_dependent_jobs[-max_parallel_jobs])
+        _dependent_jobs.append(job)
+
+    b = get_batch(name=f'Run FINEMAP on {celltypes}:{chroms}')
+    for celltype in celltypes.split(','):
+        for chrom in chroms.split(','):
```
MattWellie, line 48: It feels much simpler to move these into a config file, instead of command line arguments. That way you have a simple config file that you can pass to analysis-runner using `--config`, and the celltypes/chroms/etc. can just be recorded as a list, not a huge string that needs splitting. Centralising the settings that way makes the command line simpler, and persists those arguments in analysis-runner config.

str/fine-mapping/finemap_runner.py
```diff
@@ -0,0 +1,177 @@
...
+
+                            # Concatenate all files with data.cred* into a single file
+                            cat data.cred* > {finemap_job.ofile['data.cred']}
+
+                            # Cat data.log* into a single file
+                            cat data.log* > {finemap_job.ofile['data.log']}
+
+                            """,
+                )
+                manage_concurrency_for_job(finemap_job)
```
MattWellie, line 165: you could probably bring down time & cost by stacking multiple genes into the same image - instead of one VM per chrom * celltype * gene, you could stack ~50 genes into each job, each gene is processed in turn, then the outputs are copied out once all are completed. That would be trickier to implement, but would make better use of the available compute instead of opening a new VM for every task

str/fine-mapping/finemap_runner.py
```diff
@@ -0,0 +1,177 @@
...
+
+                finemap_job = b.new_job(name=f'Run finemap on {celltype}:{chrom} {gene}')
+                finemap_job.image(image_path('finemap'))
+                finemap_job.cpu(job_cpu)
+                finemap_job.declare_resource_group(
+                    ofile={
+                        'data.snp': '{root}.data.snp',
+                        'data.config': '{root}.data.config',
+                        'data.log': '{root}.data.log',
+                        'data.cred': '{root}.data.cred',
```
MattWellie, line 141:
```suggestion
                        'snp': '{root}.snp',
                        'config': '{root}.config',
                        'log': '{root}.log',
                        'cred': '{root}.cred',
```

partial 1/3

str/fine-mapping/finemap_runner.py
```diff
@@ -0,0 +1,177 @@
...
+                    },
+                )
+                finemap_job.command(
+                    f"""
+                            # Create a temp_file that stores the I/O file paths for FINEMAP (as per format required by the tool)
+                            temp_file=$(mktemp)
+
+                            # Write the required format to the file
+                            echo 'z;ld;snp;config;cred;log;n_samples' > $temp_file
+                            echo "{data_in.z};{data_in.ld};{finemap_job.ofile['data.snp']};{finemap_job.ofile['data.config']};data.cred;data.log;{n_samples_total}" >> $temp_file
```
MattWellie, line 151:
```suggestion
                            echo "{data_in.z};{data_in.ld};{finemap_job.ofile['snp']};{finemap_job.ofile['config']};data.cred;data.log;{n_samples_total}" >> $temp_file
```
partial 2/3

str/fine-mapping/finemap_runner.py
```diff
@@ -0,0 +1,177 @@
...
+
+                            """,
+                )
+                manage_concurrency_for_job(finemap_job)
+
+                output_path_vcf = output_path(f'finemap/ofiles/{celltype}/{chrom}/{gene}', 'analysis')
+                b.write_output(finemap_job.ofile['data.snp'], output_path_vcf + '.snp')
+                b.write_output(finemap_job.ofile['data.config'], output_path_vcf + '.config')
+                b.write_output(finemap_job.ofile['data.log'], output_path_vcf + '.log')
+                b.write_output(finemap_job.ofile['data.cred'], output_path_vcf + '.cred')
```
MattWellie, line 171:
```suggestion
                b.write_output(finemap_job.ofile, output_path_vcf)
```
partial 3/3
write_output if given a resource group will write all separate files out using their name in the declared resource group as the extension. No need to write each separately

str/fine-mapping/finemap_runner.py
```diff
@@ -0,0 +1,177 @@
...
+                            # Write the required format to the file
+                            echo 'z;ld;snp;config;cred;log;n_samples' > $temp_file
+                            echo "{data_in.z};{data_in.ld};{finemap_job.ofile['data.snp']};{finemap_job.ofile['data.config']};data.cred;data.log;{n_samples_total}" >> $temp_file
+                            chmod +x $temp_file
+
+                            # Run FINEMAP
+                            finemap --sss --in-files $temp_file --log --n-causal-snps {n_casual_snps_gene}
+
+                            # Concatenate all files with data.cred* into a single file
+                            cat data.cred* > {finemap_job.ofile['data.cred']}
```
MattWellie, line 158:
```suggestion
                            cat data.cred* > {finemap_job.ofile['cred']}
```

str/fine-mapping/finemap_runner.py
```diff
@@ -0,0 +1,177 @@
...
+
+                            # Cat data.log* into a single file
+                            cat data.log* > {finemap_job.ofile['data.log']}
```
MattWellie, line 161:
```suggestion
                            cat data.log* > {finemap_job.ofile['log']}
```

str/fine-mapping/finemap_runner.py
MattWellie: At the end of this process you write these files back out, omitting `data` from the name completely, not sure why it's being used in this resource group?

MattWellie (commented):
Some abstract thoughts on optimising VM usage,

Some specific thoughts on using an output resource group more efficiently

PR #232, opened 2024-06-16
Prepare input files for FINEMAP 

hopedisastro:
This script will prepare the z and ld files for FINEMAP, based on outputs from associaTR (meta-analysis) and corr_matrix_maker.py.

The z file contains the effect size and standard error estimates for each variant associated with a gene.
The ld file contains the correlation matrix calcualtions for each variant associated with a gene.

### DIFF
--- a/str/fine-mapping/finemap_files_prep.py
+++ b/str/fine-mapping/finemap_files_prep.py
@@ -27,9 +27,9 @@ import pandas as pd
 
 import hailtop.batch as hb
 
+from cpg_utils import to_path
 from cpg_utils.config import output_path
 from cpg_utils.hail_batch import get_batch
-from cpg_utils import to_path
 
 
 def z_file_maker(gene_name, ld_file, associatr_dir, celltype, chrom):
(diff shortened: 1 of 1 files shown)


str/fine-mapping/finemap_files_prep.py
```diff
@@ -0,0 +1,151 @@
...
+            for ld_file in ld_files:
+                gene_name = str(ld_file).split('/')[-1].split('_')[0]  # ENSG ID
+
+                if (
+                    to_path(
+                        output_path(f'finemap_prep/{celltype}/{chrom}/{gene_name}.ld', 'analysis'),
+                    ).exists()
+                    and to_path(output_path(f'finemap_prep/{celltype}/{chrom}/{gene_name}.z', 'analysis')).exists()
+                ):
+                    continue
```
MattWellie, line 135: not a biggie, but be aware of your job scaling (type * chrom * gene), and you're creating 2 existence checks per combination. All that happens in the driver job so you're delaying the start of the actual work.

I'd experiment with a change here - 

```
# get all files in the output folder, recursively, in a single query
all_files = list(to_path(output_path('finemap_prep', 'analysis')))

# check whether your intended outputs are in that list
ld_file = output_path(f'finemap_prep/{celltype}/{chrom}/{gene_name}.ld', 'analysis')
z_file = output_path(f'finemap_prep/{celltype}/{chrom}/{gene_name}.z', 'analysis')
if all (filepath in all_files for filepath in [z_file, ld_file]):
    continue
```

I thiiiiink this should scale a lot better, by posting one large query instead of thousands of individual ones.

This also builds the full output file names, so you can pass them to the relevant methods (you pass the celltype, chrom, and gene name to your methods, but you already made the full path here to check if it exists)

PR #240, opened 2024-08-29
Methylation bed concatenator

hopedisastro:
Raw methylation calls from PacBio data (using pb-CpG-tools) are provided in a BED per-sample format.

This script concatenates each per-sample BED file into a single BED file (chromosome-specific chr1-22).
Only CpG sites called in all samples are retained (inner join merge).
Only the 'mod_score' parameter is extracted; all other methylation-related columns are discarded.


michael-harper:
LGTM

Linting issues in other files were being flagged in this PR (not sure why). I've largely told the linter to ignore these issues as they are not in the scope of this PR and did not want to possibly alter functionality as I assume those scripts work already.

michael-harper (approved):
LGTM

Linting issues in other files were being flagged in this PR (not sure why). I've largely told the linter to ignore these issues as they are not in the scope of this PR and did not want to possibly alter functionality as I assume those scripts work already.

PR #242, opened 2024-09-25
Plotter for small dups in Wham 

hopedisastro:


hopedisastro:
works: https://batch.hail.populationgenomics.org.au/batches/486996

MattWellie (approved):
LGTM
there's a linting failure checking syntax, but the check seems to fail on a range of other files, rather than this one. Probably worth a PR in the near future to get those linting changes passed as a baseline for future CI

PR #244, opened 2024-10-02
File parser for meta-analysis cell type specificity approach 

hopedisastro:
This script is the first step in assessing cell type-specificity of eQTLs.
It prepares input files for the next step, which is running the meta-analysis; and also outputs a file containing eQTLs with opposite signed betas for each cell type.


annacuomo (approved):
Looks good to me, and I think you've already tested this? 
I've run black but it looks like there are a few checks failing still, although linked to other scripts?

PR #245, opened 2024-10-02
Meta-analysis of eQTL effect sizes runner script

hopedisastro:
This script performs meta-analysis of eQTLs across cell types to assess cell type-specificity of eQTLs..
Assumes that file_prep.py has been run.
This script runs R's meta package to generate pooled effect sizes for each eQTL tested between pairs of cell types.


annacuomo (approved):
LGTM! As for the other script, would be good to have a more comprehensive description of what is going on here (can go in your README under Downstream analysis), and there are a bunch of checks to fix up :)

PR #246, opened 2024-10-25
Check if shard vcf exists - skip if it does

hopedisastro:
Jobs failing: https://batch.hail.populationgenomics.org.au/batches/518758?q=&last_job_id=50 because some need more mem resources than others. 

Need to add a check to the script to check if the shard has produced a vcf; skip if it does so it doesn't start a job to create something that already exists.

### DIFF
--- a/str/runners/str_iterative_eh_runner.py
+++ b/str/runners/str_iterative_eh_runner.py
@@ -161,7 +161,7 @@ def main(
             # per sample, run parallel jobs on each shard of the catalog
             for index, subcatalog in enumerate(catalog_files, start=1):
 
-                if to_path(output_path(f'{cpg_id}/{cpg_id}_eh_shard{index}', 'analysis')).exists():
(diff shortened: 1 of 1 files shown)


str/runners/str_iterative_eh_runner.py
```diff
@@ -160,6 +160,9 @@ def main(
             )
             # per sample, run parallel jobs on each shard of the catalog
             for index, subcatalog in enumerate(catalog_files, start=1):
+
+                if to_path(output_path(f'{cpg_id}/{cpg_id}_eh_shard{index}', 'analysis')).exists():
```
MattWellie, line 164: You should save this string as a variable and reuse below

MattWellie (approved):

str/runners/str_iterative_eh_runner.py
hopedisastro: It's a master path; I should check for .vcf specifically.

PR #247, opened 2024-11-03
Rolling merge version of mergeSTR runner

hopedisastro:
I will still keep the merge_str_runner.py script but add this to the repo for rolling merges. 

rolling_merge_str_runner.py will merge multiple mergeSTR outputs together into a new mergeSTR VCF. 

This gets around scaling issues with Hail Batch which errorrs out when too many file input params are specified
https://centrepopgen.slack.com/archives/C030X7WGFCL/p1730386942089209


hopedisastro:
seems to be ok; all shards expected to fail except for shard 27 in `test`: https://batch.hail.populationgenomics.org.au/batches/529076

MattWellie (approved):
Sorry for the delay, I wanted to tidy up some of the code but it's not important. As always, something came up 😅

PR #252, opened 2024-11-22
Remove PID parsing

hopedisastro:
Not really needed.

PID parsing was tripping up the new set of VCFs for 10k10k p1, so I'm going going to disable it. https://batch.hail.populationgenomics.org.au/batches/530692


annacuomo (approved):

annacuomo:
needs linting but otherwise looks good to me!

PR #254, opened 2024-11-26
Use sample id file to read in sample IDs 

hopedisastro:
Adding very many sample IDs as a command line argument is not a good idea; we convert the input as a command line argument into a sample id file that gets read in (similar to the EH script).


str/runners/str_iterative_hipstr_runner_sharded.py
```diff
@@ -78,14 +78,14 @@ def get_cloudfuse_paths(dataset, input_cpg_sids):
 
 
 # inputs:
+@click.option('--sample-id-file', help='List of CPG IDs')
```
hopedisastro, line 81:
```suggestion
@click.option('--sample-id-file', help='CSV of CPG IDs')
```

str/runners/str_iterative_hipstr_runner_sharded.py
```diff
@@ -78,14 +78,14 @@ def get_cloudfuse_paths(dataset, input_cpg_sids):
 
 
 # inputs:
+@click.option('--sample-id-file', help='CSV of CPG IDs')
```
hopedisastro, line 81:
```suggestion
@click.option('--sample-id-file', help='Path to CSV of CPG IDs')
```

MattWellie:
This would be even easier if you load the list of IDs into the config file (assuming you're using analysis-runner). There would be no need to push a file to GCP, and there would be a permanent record of the final config (a sample ID file could be overwritten)

PR #255, opened 2024-11-26
add.split(,) for config param in HipSTR script

hopedisastro:


MattWellie (changes requested):
Please don't use this approach, just add a list to the config file instead of a single string

PR #257, opened 2024-12-19
use "None" because TOML doesn't accept None

hopedisastro:
https://centrepopgen.slack.com/archives/C030X7WGFCL/p1716521149097559


katiedelange (approved):
Just noting the changes here that I am approving, to make sure they are what was intended:
- No longer able to remove specific samples (this means we cannot explicitly exclude the contaminated sample, but this should get dropped due to missing scRNA-seq data anyway)
- No longer able to condition on specific SNP genotypes

PR #263, opened 2025-05-09
Trans eQTL runner

htanudisastro:
Remove --regions so that it runs all variants in the vcf.

### DIFF
--- a/str/trans-eqtl/trans_associatr_runner.py
+++ b/str/trans-eqtl/trans_associatr_runner.py
@@ -1,0 +1,101 @@
+#!/usr/bin/env python3
+# pylint: disable=missing-function-docstring,no-member,too-many-arguments
+"""
+This script runs associaTR, given a chromosome or cell type.
+Ensure prior scripts have been run to generate dependent files, particularly:
+- get_cis_numpy_files.py
+- pseudobulk.py
+- get_covariates.py
+- qc_filters_associatr.py (depends on qc_annotator.py)
+
+ analysis-runner --dataset "bioheart" --config associatr_runner_trans.toml \
+    --description "run associatr" \
+    --access-level "full" \
+    --memory "8G" \
+    --output-dir "str/associatr/trans_pilot/bioheart_n975" \
+     python3 trans_associatr_runner.py
+
+
+"""
+import json
+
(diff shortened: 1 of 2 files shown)


htanudisastro:
Is there a way to bypass the Ubuntu image check or do I have to update the image?

str/trans-eqtl/trans_associatr_runner.py
```diff
@@ -0,0 +1,99 @@
...
+from cpg_utils.config import get_config
+from cpg_utils.hail_batch import get_batch, init_batch, output_path
+
+
+def main():
+    """
+    Run associaTR processing pipeline
+    """
+    b = get_batch(name='Run associatr')
+    init_batch()
```
MattWellie, line 36: I don't think this is needed here? `init_batch` starts a hail query runtime, but you only need a batch

str/trans-eqtl/trans_associatr_runner.py
```diff
@@ -0,0 +1,99 @@
...
+        _dependent_jobs.append(job)
+
+    for celltype in get_config()['associatr']['celltypes'].split(','):
+        for chromosome in get_config()['associatr']['chromosomes'].split(','):
+            vcf_file_path = get_config()['associatr']['vcf_file_path']
+            variant_vcf = b.read_input_group(
+                **dict(
+                    base=vcf_file_path,
+                ),
+            )
```
MattWellie, line 56: this can be moved outside the loops, the VCF isn't specific to celltypes or chromosomes

str/trans-eqtl/trans_associatr_runner.py
```diff
@@ -0,0 +1,99 @@
...
+
+"""
+import json
+
+import pandas as pd
+
+import hailtop.batch as hb
+
+from cpg_utils import to_path
+from cpg_utils.config import get_config
```
MattWellie, line 27: you might want to use `config_retrieve`, the syntax is a little cleaner, and has built-in informative error messaging if you try and access keys which don't exist, e.g.

```
get_config()['associatr']['celltypes']
vs.
config_retrieve(['associatr', 'celltypes'])
```

PR #268, opened 2025-06-06
Change index for which pvalue is extracted from. 

htanudisastro:
Meta-analysis output file now has pvalue index at position 12.


silkm:
Feel like just adding a sanity check message since your PR description doesn't mention, I assume you're definitely intending to change the project from `bioheart-main` to `tenk10k-test` here for your inputs

PR #270, opened 2025-06-06
Stylistic changes to QQPlotter

htanudisastro:
- Fix the x/y axes to match obs/expected log p value labels. 
- Add color mapping in concordance with TenK10K P1 color scheme. 
- Enforce cell types being listed in the same order as specified in command line args.


silkm (approved):
LGTM

PR #271, opened 2025-06-09
MashR preparation scripts 

htanudisastro:
MashR is a method that calculates the pairwise similarity of tissues/celltypes while taking into account differences in power between cell types. 

Importantly, mashR can only compare eQTLs that are found in all 28 cell types (input matrix cannot have missing values for eQTLs tested in one cell type but not another) "strong effects model", and also uses a null model of significant and non-significant tests as a baseline for its shrinkage model. 

These two scripts prepare the input files for mashR. 

Specifically

Prep_inputs.py extracts the corresponding effect sizes and standard errors in the other 27 cell types for each eQTL that passed FDR <5% in one cell type (this is the starting base for inputs into the "strong effects model"). It also extracts the effect sizes and standard errors of all variants tested across all 28 cell types for a specific chromosome (these will be used for the null model input). 

Concatenate_inputs.py first finds the intersection of variant/gene pairs tested in all 28 cell types for the strong effects model; then it performs a columnwise concatenation; where rows represent a variant/gene pair and columns consist of celltype_beta, celltype_se for all the 28 cell types. 
Similar procedure performed for the null effects model input.


htanudisastro:
This is not a blocker. Does not need a thorough review; results look as expected (to generate a heatmap essentially for a figure in the paper).

silkm (approved):

PR #274, opened 2025-06-10
Cell spec cuomo master

htanudisastro:


str/cell-type-spec/master_cell_spec_runner.py
```diff
@@ -0,0 +1,308 @@
...
+    row_cell_type, scenario_2_celltypes, egene_chrom, row_pos, row_end, egene, row_motif, meta_scen2_path
+):
+    """
+    Reprocesses scenario 2 to check if can be reclassified as scenario 4/5 based on meta-analysis p value, based on the criteria from Cuomo et al. 2025.
+
+    """
+    reprocess_scen2_dict = {}
+
+    # Load the meta results for the given cell type
+    meta_results = pd.read_csv(f'{meta_scen2_path}/{row_cell_type}/meta_results.tsv', sep='\t')
```
folded, line 123: Assuming the tsvs are not too large to fit into memory at the same time, I'd suggest extracting this into a cached function to avoid reloading each file multiple times:

```
@functools.cache
def _load_meta_results(path: str, cell_type: str) -> pd.DataFrame:
    return pd.read_csv(f'{path}/{cell_type}/meta_results.tsv', sep='\t')

...
    meta_results = _load_meta_results(meta_scen2_path, row_cell_type)
...
```

str/cell-type-spec/master_cell_spec_runner.py
```diff
@@ -0,0 +1,308 @@
...
+def process_scenario_1_2(gene, chrom, cell_types_without_egene, associatr_path):
+    """
+
+    Processes scenario 1 and 2
+    """
+    scenario_1_2_dict = {}
+    for cell_type in cell_types_without_egene:
+        try:
+            df = pd.read_csv(f'{associatr_path}/{cell_type}/{chrom}/{gene}_100000bp_meta_results.tsv', sep='\t')
+            scenario_1_2_dict[cell_type] = 2
```
folded, line 25: Are `1` and `2` meaningful? It might be good to define an enum for them to make it clear what they mean.

str/cell-type-spec/master_cell_spec_runner.py
```diff
@@ -0,0 +1,308 @@
...
+
+def process_scenario_1_2(gene, chrom, cell_types_without_egene, associatr_path):
+    """
+
+    Processes scenario 1 and 2
+    """
+    scenario_1_2_dict = {}
+    for cell_type in cell_types_without_egene:
+        try:
+            df = pd.read_csv(f'{associatr_path}/{cell_type}/{chrom}/{gene}_100000bp_meta_results.tsv', sep='\t')
```
folded, line 24: Reading the csv and not using the values is inefficient, especially as this involves making a network request to GCS. You could test for existence of the file instead, assuming that you're not concerned about whether the file parses correctly given that it exists.

Also, if this might be called multiple times for a given gene, then it would be worth caching the result.

PR #278, opened 2025-06-18
STEP 1 of Finemap MCV workflow - get genotypes (dosages) 

htanudisastro:
This is the first step of the finemapping workflow. 

For every gene associated with an eTR (TR eQTL): 

- we extract the cis window (searchspace of variants), which is +/- 100kb of the gene body. 
- then we call on helper functions `tr_extract_genotype_matrix` and `snp_extract_genotype_matrix` to extract the GTs of TR and SNP variants respectively in the cis window. 
- the TR and SNP genotype matrices (or DFs) are merged, and outputted. 
- GT output files will be used in the next step of the workflow. 

Pretty confident this works OK.

### DIFF
--- a/str/fine-mapping/mcv/files_prep_dosages.py
+++ b/str/fine-mapping/mcv/files_prep_dosages.py
@@ -168,13 +168,12 @@ def main(estrs_path):
     df = df.drop_duplicates(subset=['gene_name', 'chr'])
     # sort by chromosome
     for chrom in df['chr'].unique():
-        if chrom!= 'chr1':
-            df_chr = df[df['chr'] == chrom]
+        df_chr = df[df['chr'] == chrom]
 
-            j = b.new_python_job(name=f'Prepare for {chrom}')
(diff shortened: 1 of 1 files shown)


str/fine-mapping/mcv/files_prep_dosages.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+@click.option('--estrs-path', type=str, required=True, help='Path to the estrs file.')
+@click.command()
+def main(estrs_path):
+    b = get_batch(name='Dosages files prep for SuSie MCV')
+    df = pd.read_csv(estrs_path)
+    df = df.drop_duplicates(subset=['gene_name', 'chr'])
+    # sort by chromosome
+    for chrom in ['chr8']:
```
silkm, line 170: I know / assume you'll remove this chr filter for the full run, just flagging so I don't forget

silkm (approved):
LGTM

PR #301, opened 2026-07-24
Create SV VCF file for input into associaTR

htanudisastro:
This script converts an existing GATK-SV VCF file into a
mock ExpansionHunter-style VCF for direct use with associaTR.

The objective is to: 
- extract the GTs from GATK-SV and list it in a EH-style variant record

Please note ad-hoc changes:
- RU field stores the REF and ALT alleles in the format of 'REF-ALT' (REF/ALT
  here are the SV's symbolic alleles, e.g. 'N-<DEL>', not sequence bases)

- RL field is set to 0 for all loci so that the REF allele is coded as 0
  (https://github.com/gymrek-lab/TRTools/blob/master/trtools/utils/tr_harmonizer.py#L515)
- REF field (INFO) is set to 3 (arbitrary filler value)
- 
- SVTYPE and END are taken from the original GATK-SV INFO field,
  since GATK-SV already provides these and they're meaningful for SVs

- GT is extracted from whichever FORMAT slot it occupies (GATK-SV FORMAT is
  typically GT:ECN:EV:GQ:OGQ:RD_CN:RD_GQ:SL), not assumed to be field 0 only
  based on position -- it's located by name in the FORMAT header each time

- Multi-allelic CNVs (FILTER contains MULTIALLELIC) do not carry meaningful
  GT calls in GATK-SV; for these records we read FORMAT/CN (copy number) per
  sample and emit it as "CN/0" in the mock GT slot. associaTR sums the two
  alleles per sample, so effective per-sample dosage = CN. Missing CN is
  written as "./.". The monomorphic-site drop uses CN directly for these
  records.

I have proofed the outputs and they look reasonable.

### DIFF
--- a/tenk10k-sv/sv_vcf_for_associatr.py
+++ b/tenk10k-sv/sv_vcf_for_associatr.py
@@ -1,0 +1,224 @@
+#!/usr/bin/env python3
+"""
+This script converts an existing GATK-SV VCF file (chromosome-specific) into a
+mock ExpansionHunter-style VCF for direct use with associaTR.
+
+Please note ad-hoc changes:
+- RU field stores the REF and ALT alleles in the format of 'REF-ALT' (REF/ALT
+  here are the SV's symbolic alleles, e.g. 'N-<DEL>', not sequence bases)
+- RL field is set to 0 for all loci so that the REF allele is coded as 0
+  (https://github.com/gymrek-lab/TRTools/blob/master/trtools/utils/tr_harmonizer.py#L515)
+- REF field (INFO) is set to 3 (arbitrary filler value, matches SNP version)
+- SVTYPE and END are taken from the original GATK-SV INFO field (not faked),
+  since GATK-SV already provides these and they're meaningful for SVs
+- GT is extracted from whichever FORMAT slot it occupies (GATK-SV FORMAT is
+  typically GT:ECN:EV:GQ:OGQ:RD_CN:RD_GQ:SL), not assumed to be field 0 only
+  based on position -- it's located by name in the FORMAT header each time
+- Multi-allelic CNVs (FILTER contains MULTIALLELIC) do not carry meaningful
+  GT calls in GATK-SV; for these records we read FORMAT/CN (copy number) per
+  sample and emit it as "CN/0" in the mock GT slot. associaTR sums the two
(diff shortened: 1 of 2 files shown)


tenk10k-sv/sv_vcf_for_associatr.py
```diff
@@ -0,0 +1,276 @@
...
+@click.option('--vcf-path', required=True, help='Input VCF file')
+@click.option('--job-storage', default='20G')
+@click.option('--job-cpu', default=1)
+@click.command()
+def main(vcf_path, job_storage, job_cpu):
+    b = get_batch(name='SV VCF maker for associaTR')
+
+    sv_vcf = vcf_path
+    output_file = output_path(
+        f'sv_vcf_for_associatr/{os.path.basename(vcf_path)}',
```
silkm, line 322: Walking through temp output to final output with some claude, the temp file is written uncompressed which means the final output file has a `.gz` extension but is also uncompressed. So would want to set the temp file to being output compressed I think?

tenk10k-sv/sv_vcf_for_associatr.py
```diff
@@ -0,0 +1,276 @@
...
+    n_seen = 0
+    n_written = 0
+
+    with gzip.open(vcf_file, 'rt') as fin, open('temporary_gt_file.txt', 'w') as fout:
+        for line in fin:
+            if line.startswith('#CHROM'):
+                # Insert the mock ExpansionHunter-style header fields right before
+                # the #CHROM line (GATK-SV VCFs don't have a ##hailversion line to
+                # key off of, unlike the hail-derived SNP VCFs).
+                fout.writelines(NEW_FORMAT_FIELDS)
```
silkm, line 145: Claude suggestion: a few header lines may end up duplicated for END/SVTYPE/GT from bringing the original GATK-SV headers

htanudisastro: implemented!

tenk10k-sv/sv_vcf_for_associatr.py
htanudisastro: implemented compressed gz file!

silkm (approved):
LGTM!

PR #302, opened 2026-08-13
Identify proximal ATAC peaks

htanudisastro:
## Add SV-proximal ATAC peak filtering

Identifies the ATAC-seq peaks that fall within a window (default 1Mb) of any SV in a
given VCF, per cell type. The output slots into the current ATAC
associatr path.

### Why

`str/atac/get_cis_numpy.py` filters peaks down to a precomputed "sites within 1Mb of an
eSV" list read from `loc_filtered_dir` (`{cell_type}_loc.csv`). There was no equivalent
for SVs, so there was no way to restrict the ATAC association testing to peaks near a
structural variant.

### What it does

Two stages, both in one Hail Batch:

1. **`Extract SV intervals`** (bcftools, runs once) — pulls `CHROM/POS/INFO/END` out of the
   VCF into a compact TSV, also written to the output dir as `sv_intervals.tsv` for
   provenance. Doing this once keeps the multi-GB VCF off the analysis-runner driver and
   out of the per-cell-type jobs.
2. **`Identify SV-proximal peaks`** (one Python job per cell type) — reads just the `loc`
   column of `PseudobulkMatStandardised.csv`, expands each SV to
   `[POS - window_size, END + window_size]`, merges those into disjoint intervals, and
   tests each peak with a single binary search.

### Output

`{cell_type}_loc.csv`, a single `loc` column — a drop-in for the `loc_filtered_dir` input
of `str/atac/get_cis_numpy.py`, so switching that pipeline from eSTR-proximal to
SV-proximal peaks is a config change only.

### DIFF
--- a/tenk10k-sv/atac/get_sv_proximal_peaks.py
+++ b/tenk10k-sv/atac/get_sv_proximal_peaks.py
@@ -1,0 +1,143 @@
+#!/usr/bin/env python3
+# pylint: disable=too-many-locals
+
+"""
+This script aims to:
+ - identify the ATAC-seq peaks ('loc' column of PseudobulkMatStandardised.csv) that fall within
+   a window (default 1Mb) of any SV in a VCF, per cell type.
+ - output {cell_type}_loc.csv (a single 'loc' column), which is the format expected by the
+   loc_filtered_dir input of str/atac/get_cis_numpy.py.
+
+SV coordinates are taken as POS to INFO/END; SVs without an END (eg BND, INS) are treated as
+POS to POS. A peak is retained if it overlaps [POS - window_size, END + window_size].
+
+analysis-runner --config get_sv_proximal_peaks.toml --dataset "bioheart" --access-level "test" \
+--description "get SV-proximal ATAC peaks" --output-dir "str/atac-seq/sv-proximal-peaks/1mb" \
+python3 get_sv_proximal_peaks.py
+
+"""
+
+from cpg_utils.config import get_config
+from cpg_utils.hail_batch import get_batch, image_path, output_path
+
+
+def sv_proximal_loc_extractor(
+    sv_intervals_file,
+    pseudobulk_dir,
+    cell_type,
+    window_size,
+):
+    """
+    Writes the subset of peaks lying within window_size of an SV to {cell_type}_loc.csv
+
+    """
+    import numpy as np
+    import pandas as pd
+
+    from cpg_utils.hail_batch import output_path
+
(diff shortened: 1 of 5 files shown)


tenk10k-sv/atac/get_sv_proximal_peaks.toml
```diff
@@ -0,0 +1,13 @@
+[get_sv_proximal_peaks]
+pseudobulk_dir = 'gs://cpg-bioheart-test/str/atac-seq'
+sv_vcf_path = 'gs://cpg-tenk10k-test/pub-analysis/final-vcf/filtered/tob_common_maf_gte_1pct.vcf.gz'  # TODO path to the SV VCF (bgzipped)
```
joshuamschmidt: This path exists, so the job will run on it either way. Can you confirm it's the VCF you want and drop the TODO? There's also `gs://cpg-tenk10k-sv-test/pub-analysis/final-vcf/filtered/` (the `tenk10k-sv` dataset, where #301 writes) which is easy to confuse with this one.

tenk10k-sv/atac/get_sv_proximal_peaks.py
```diff
@@ -0,0 +1,143 @@
...
+
+"""
+This script aims to:
+ - identify the ATAC-seq peaks ('loc' column of PseudobulkMatStandardised.csv) that fall within
+   a window (default 1Mb) of any SV in a VCF, per cell type.
+ - output {cell_type}_loc.csv (a single 'loc' column), which is the format expected by the
+   loc_filtered_dir input of str/atac/get_cis_numpy.py.
+
+SV coordinates are taken as POS to INFO/END; SVs without an END (eg BND, INS) are treated as
+POS to POS. A peak is retained if it overlaps [POS - window_size, END + window_size].
```
joshuamschmidt, line 12: Worth a line on why the window covers the whole SV span rather than just the breakpoints. I think span + flank is right (a peak inside a large DEL is worth testing), but it does mean a few big CNVs can move the retained peak count a lot.

joshuamschmidt (commented):
Had a good look at this, 

A few small things inline. Branch is 2 behind main if you want to rebase first.

tenk10k-sv/atac/get_sv_proximal_peaks.py
htanudisastro: thanks Josh, comments all addressed!

joshuamschmidt (approved):
LGTM

